PR #374, opened 2025-06-12
update cost allocation screenshots

jpvotta:


kumare3 (approved):
LGTM, but just ensuring we checkin pngs into the repo like so - cc @nelsonjr

jpvotta:
FWIW @kumare3 these are replacing old pngs of the same name.

PR #417, opened 2025-07-22
Add concurrency control docs to launch plan section

thomasjhuang:
Add concurrency control docs per this flyte PR - https://github.com/flyteorg/flyte/pull/6475. Related flytekit change - https://github.com/flyteorg/flytekit/pull/3267


ppiegaze (changes requested):
Please target the `v1` branch, not `main`. This is a v1 feature

thomasjhuang:
@ppiegaze Hi there, I've updated to the `v1` branch. Also I do think we should check in the flytekit change before this goes in.

PR #418, opened 2025-07-23
[Docs] Add warning block for installing ca certificate for specific usage

machichima:
As discussed in this issue: https://github.com/flyteorg/flyte/issues/6438, it is unclear that Flyte `ImageSpec`'s default image does not include apt package `ca-certificate`, which can cause confusion when user want to use polars.

Document this in `ImageSpec` docs


machichima:
Hi @ppiegaze , I changed the target branch to v1. Thank you!

PR #422, opened 2025-07-29
Add banner to v2

ppiegaze:


ppiegaze:
colors,font,etc,incorrect
check csss

PR #434, opened 2025-08-01
add deep research tutorial

samhita-alla:


layouts/shortcodes/code.html
```diff
@@ -58,7 +58,7 @@
   {{- end -}}
   {{- if not $foundStart -}}
     {{- warnf "Fragment marker '%s' not found in file %s. Please mark a section with the following markers: %s and %s"
-          $fragment $fileName (printf "{{docs-fragment %s}}" $fragment) "{{/docs-fragment}}" -}}
+          $fragment $fileName (printf "{{docs-fragment %s}}" $fragment) (printf "{{/docs-fragment %s}}" $fragment) -}}
```
nelsonjr, line 61: This does not look correct. The closing that should be always `{{ /docs-fragment }}`

nelsonjr, line 61: This does not look correct. The closing that should be always {{ /docs-fragment }}. Revert this line?

nelsonjr (changes requested):

PR #505, opened 2025-09-10
[WIP] V1 SEO Improvements

ursucarina:
see detailed seo improvements doc: https://www.notion.so/Union-ai-Docs-SEO-UX-Improvements-26a8cc06513d80a4921ec4a849eb4d82

** Note: use this [chrome extension](https://chromewebstore.google.com/detail/ograph-previewer/ggcfeakcnodgcmmllfdbmngekljbhiim) to visualize how the added meta tags are going to be used when sharing across social media platforms
image
image

### DIFF
--- a/themes/union/layouts/partials/sidebar.html
+++ b/themes/union/layouts/partials/sidebar.html
@@ -61,7 +61,8 @@
 {{- end -}}
 
 {{ define "site-nav" }}
-<nav class="site-nav">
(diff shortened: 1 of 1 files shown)


ppiegaze:
I have changed the logic so that the description used in meta tags is generated from the original HTML content (Hugo `.Content` context variable) instead of from the pre-generated `.Summary` context variable provided by Hugo (and already stipped of HTML tags). This allows us to intelligently remove the errant H1 tags from the description, whereas using the >summary is more difficult since the H1 info was already stripped

PR #885, opened 2026-04-02
Update External Authz docs for ExternalIdentity (FAB-189)

mhotan:
## Summary

- Update identity resolution docs to document all three proto types (`external_identity`, `user_id`, `application_id`) and when each is sent
- Clarify that `identitytype` custom claim is **not required** for External authorization
- Add token availability table (not every request carries a token — background workers don't)
- Update reference implementation to extract subject from whichever proto type arrives
- Document that servers should match subject against known service account IDs, not rely on proto type

## Context

FAB-189: When the `identitytype` JWT claim is absent (selfhosted with non-Okta IdPs), the platform now sends `ExternalIdentity{subject}` instead of failing. This PR updates the selfhosted authorization docs to reflect the new behavior.

Companion PRs:
- cloud [#15185](https://github.com/unionai/cloud/pull/15185) — ExternalIdentity code fix
- Notion analysis: https://www.notion.so/3368cc06513d815dbeebee18d0216c07

Closes #880 (superseded).

ref FAB-189

### DIFF
--- a/content/deployment/selfhosted-deployment/authorization.md
+++ b/content/deployment/selfhosted-deployment/authorization.md
@@ -145,7 +145,7 @@ service AuthorizerService {
 
 | Field | Type | Description |
 |-------|------|-------------|
-| `identity` | `Identity` | The caller — an `external_identity` containing the subject string and the raw OIDC token (when available) |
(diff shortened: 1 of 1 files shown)


ppiegaze:
Superseded by #935, which adapts this content to the new flyte/union variant structure (stacked on #934).

PR #886, opened 2026-04-02
Notification docs

iaroslav-ciupin:
* Running a single run with notifications
* Attach notifications to a trigger


ppiegaze (commented):
The changes under content/api-reference/flyte-sdk/ should not be made here. These files are generated from flyte-sdk source. Theses changes should therefore be made in the docstrings of flyte-sdk

PR #937, opened 2026-04-22
Guide for cicd in v2

dansola:


kumare3 (approved):

ppiegaze:
Fixed the broken links (`../monorepo-with-uv` → `./monorepo-with-uv`) and changed variant to `-flyte +union` (API keys and `flyteplugins-union` are Union-specific).

One question for @dansola: Lines 43 and 67-80 seem contradictory about where `org` comes from. Line 43 says the API key format is `endpoint:client_id:client_secret:org` (org embedded in the key), but line 80 says `FLYTE_API_KEY` overrides only the auth fields and `org` comes from `config.yaml`. Which is correct? Does the key contain `org` but config.yaml can override it, or does the key not actually carry `org`?

dansola:
> Fixed the broken links (`../monorepo-with-uv` → `./monorepo-with-uv`) and changed variant to `-flyte +union` (API keys and `flyteplugins-union` are Union-specific).
> 
> One question for @dansola: Lines 43 and 67-80 seem contradictory about where `org` comes from. Line 43 says the API key format is `endpoint:client_id:client_secret:org` (org embedded in the key), but line 80 says `FLYTE_API_KEY` overrides only the auth fields and `org` comes from `config.yaml`. Which is correct? Does the key contain `org` but config.yaml can override it, or does the key not actually carry `org`?

@ppiegaze . good point. i removed this to remove confusion

PR #550, opened 2025-10-31
add docs on self managed plugins

dansola:
tested internally on the union `arn:aws:eks:us-east-1:356633062068:cluster/oc-canary-daniel-byok` cluster

PR #943, opened 2026-04-23
Add compute plugins configuration for self-managed deployment

ppiegaze:
## Summary

- Add new page documenting how to install and configure Dask and Ray K8s operators on a self-managed data plane
- Covers operator installation (Helm), plugin enablement, and log/dashboard link template configuration
- Tabbed AWS/GCP sections for cloud-specific log configuration

Based on content from #550 (which is stale and incompatible with the current site structure). Closes #550.

## Test plan

- [x] `make dist` builds cleanly
- [ ] Page appears under Self-managed deployment > Configuration > Compute plugins
- [ ] Tabs switch correctly between AWS and GCP config
- [ ] Only appears in union variant


ppiegaze:
@dansola. I adapted this PR from the old #550 to fit the new content structure. There were some comments on that one which still apply. PTAL

dansola (approved):

PR #953, opened 2026-04-27
Port GCP E2E to Flyte 2 + dataplane image SHA override

EngHabu:
## Summary

- **Port `selfmanaged_gcp_e2e.py` to the Flyte 2 function-style pattern** that AWS already uses. Replaces the synchronous/uctl-driven module with async tasks, frozen per-resource records (`GKECluster`/`GCSBucket`/`ARRepo`/`ServiceAccount`), a `GCPResources` aggregate, and a phased `main` that fans out infra creation under `flyte.group(...)`. Auto-resolves `project_id` from `gcloud config` when `--project_id` is omitted. Buildkit pool creation is now idempotent on its own (separate from cluster creation) with `--num-nodes 1` / `--min-nodes 1` so pods schedule immediately. Local-only credential hydration; remote-launch raises a clear "not yet implemented" since the shared bootstrap secret only carries AWS STS today.
- **Add `dataplane_image_sha` override.** Plumbed through `BaseConfig`, `helm_install` (`--set image.union.tag=<sha>`), and both `main` / `launch_remote` for AWS and GCP.
- **Auth/credential fallbacks.** `init_union_client` now falls back from `UNION_API_KEY` → `FLYTE_API_KEY`. GCP resolves `project_id` from `gcloud config get-value project` when not passed.
- **Helm robustness fixes** (`selfmanaged_common.py`):
  - `helm repo add unionai ... --force-update` so a stale local entry is replaced rather than erroring.
  - `helm repo update unionai` (scoped) so an unrelated broken repo on the user's local helm config doesn't kill the run.
- **Smoke test fixes** (`smoke_tests.py`):
  - `run.sync()` instead of `await run.sync.aio()` (the sync method isn't `_SyncWrapper`-wrapped, so it has no `.aio`).
  - Add `fastapi` + `flyteplugins-union` to the image-builder / image-cache / reuse / app-deploy test images.
- **Debug-friendly defaults**: `logging.basicConfig(level=DEBUG)` across e2e modules; `cache=\"disable\"` on the AWS env (matching GCP) so reruns don't replay stale outputs; `flyte[tui]` added to `pyproject.toml`.

## Test plan

- [ ] `flyte run --local selfmanaged_gcp_e2e.py main --control_plane_url ... --cluster_name ... --skip_teardown True` — verify Phases 1–3 complete on a fresh GCP project.
- [ ] Same but with `--dataplane_image_sha <sha>`; check the helm release uses the override (`kubectl -n union get deploy union-operator -o yaml | grep image:`).
- [ ] AWS regression: `flyte run --local selfmanaged_aws_e2e.py main ...` still passes.
- [ ] Smoke suite: `_assert_succeeded` no longer errors with `'function' object has no attribute 'aio'`.

## Notes / follow-ups

- `BaseConfig.helm_chart_branch` defaults to `enghabu/skip-registries` in this PR — a WIP branch used during validation. Reset to `\"\"` (published chart) before merge.
- GCP remote launch needs the shared bootstrap secret to carry a GCP service-account JSON; tracked separately.
- Knative `controller` KSA WI binding + per-project task-pod KSA bindings on GCP are not in this PR — see audit notes.


ppiegaze:
@Haytham. Can you resolve the conflicts?

PR #956, opened 2026-04-27
Add CoreWeave to self-managed deployment guides

ppiegaze:
## Summary

- Adds a CoreWeave provider page at `content/deployment/selfmanaged/selfmanaged-coreweave/_index.md` and links it from the self-managed landing page (alongside AWS, GCP, Azure, OCI, Generic).
- Content sourced from CoreWeave's [Deploy Union.ai on CKS](https://docs.coreweave.com/products/cks/clusters/frameworks/union-ai) guide.

## Adaptations from the source

- Frontmatter and H1 follow our IA convention (`Data plane setup on CoreWeave`, `weight: 7`, `variants: -flyte +union`).
- Mintlify `<Note>` / `<Warning>` → GitHub-style `> [!NOTE]` / `> [!WARNING]`.
- Stripped `theme={"system"}` from code fences.
- Internal CoreWeave links rewritten to absolute `https://docs.coreweave.com/...` URLs.
- `flyte` and `uctl` CLI links repointed to v2 docs (`../../../api-reference/...`) instead of 1.x sources.
- Endpoint switched from `cloud-staging.union.ai` to prod `union.ai`.
- Dropped the "Register the cluster" section that walked the customer through cloning Union's internal `cloud` repo, running `make generate_tenant_configs`, and triggering a Buildkite build — that's Union ops tooling, not customer-facing. Kept the `uctl get cluster` verification and folded it into the deploy section as the final step (matches the AWS guide's pattern).
- Fixed a malformed troubleshooting table row (a `kubectl ... | grep` cell broke column count) by escaping the pipe.

## Test plan

- [ ] `make dist` builds clean
- [ ] `make check-links` passes
- [ ] Sidebar shows CoreWeave under self-managed
- [ ] Visual check of admonitions, code blocks, troubleshooting table


ppiegaze:
@Eunwoo (or maybe @mhotan)
Can you address the issue flagged by claude and copilot. Claude has summarized it here:
 
## Re: Copilot comment on the Helm install method (line 233)

Copilot flagged that this guide installs charts via `git clone unionai/helm-charts` + tag checkout, while other self-managed guides (AWS/GCP/etc.) use the published Helm repo. Suggested swap:

```
helm repo add unionai https://unionai.github.io/helm-charts/
helm install unionai/dataplane-crds
helm install unionai/dataplane
```

I verified against `https://unionai.github.io/helm-charts/index.yaml` and the latest `dataplane-2026.4.9.tgz` before deciding what to do. Findings:

### What's actually true

1. **Both charts are published.**
   - `unionai/dataplane` — latest `2026.4.9`
   - `unionai/dataplane-crds` — latest `2026.4.6`
   - Independent release cadences — no shared version tag.

2. **CRDs are NOT bundled in `unionai/dataplane`.** The latest tarball has no `crds/` and no `templates/crds/` directory, and `dataplane-crds` is not in `Chart.yaml` dependencies. A separate `unionai/dataplane-crds` install is required.

3. **The current AWS guide (`selfmanaged-aws/deploy-dataplane.md`) is missing the CRDs install step.** It runs only `helm upgrade --install union unionai/dataplane`. That's a real gap in the AWS guide, independent of this PR.

4. **`provider: metal` is inert** in the published chart — `_helpers.tpl` only branches on `aws`/`gcp`/`gcs`/`azure`. The CoreWeave config relies on `storage.provider: custom` (which IS supported), so swapping install methods doesn't change correctness.

5. **`helm dependency build` is only needed for unpacked git chart sources.** Published tarballs already vendor deps under `charts/`. That step disappears with the helm-repo install.

### Equivalent fix

The two-step install (NOT the one-step AWS pattern, which would silently drop CRDs):

```bash
helm repo add unionai https://unionai.github.io/helm-charts/
helm repo update

helm upgrade --install unionai-dataplane-crds unionai/dataplane-crds

helm upgrade --install unionai-dataplane unionai/dataplane \
  --namespace union --create-namespace \
  --values <PATH-TO-VALUES-FILE> \
  --timeout 10m
```

### Decisions before applying the fix

1. **Pin versions or follow latest?**
   - Current PR pins via git tag → both charts pinned in lockstep, fully reproducible.
   - AWS guide pins nothing → latest-wins.
   - Helm-repo equivalent of pinning needs two `--version` flags, and since `dataplane-crds` and `dataplane` release on different cadences, the user has to know which pair is compatible.
   - **Open question:** is there a documented compatibility matrix? If not, dropping the pin and matching the AWS guide is simpler but less reproducible.

2. **Fix the AWS-guide CRDs gap as a separate follow-up?**
   - Confirmed missing. Worth a separate PR (and probably affects GCP/Azure/OCI guides too — haven't checked yet).

3. **Should this CoreWeave guide be split into the `prepare-infra.md` + `deploy-dataplane.md` pattern** (Copilot's structural comment on line 11) before we touch the install method? If we're going to refactor anyway, do both at once.

Holding off on the edit until these are resolved.

PR #997, opened 2026-05-11
Fix data-retention page: metadata bucket holds inputs/outputs, not DB state

ppiegaze:
## Summary

- Rewrites the "How the platform uses the default bucket" section so it accurately describes what's in the metadata bucket: task **inputs and outputs** (serialized Protobuf literals passed between tasks), plus pointers to offloaded raw data — **not** control plane database state. Ketan flagged the current wording as a docs bug in [Slack](https://unionai.slack.com/archives/C07QTRXMH8C/p1778239756099139): "no this is not how flyte os works. this is a bug in docs. metadata should be inputs and outputs."
- Adds a callout distinguishing the object-store "metadata" path from control plane DB state (workflow registrations, execution status, schedules, audit history — none of which are subject to bucket lifecycle policies).
- Tightens the "Impact of metadata loss" table so the UI/operations rows distinguish what's actually broken by metadata loss (literal previews, deck data, artifact payloads) versus what's served from the DB (execution list, run-when audit log).
- Scopes the existing "Customizing the metadata path" Helm guidance to **self-managed** deployments.
- Adds a "Per-run customization" section pointing to [`flyte.with_runcontext()`](https://www.union.ai/docs/v2/union/user-guide/task-deployment/run-context/#storage) (`run_base_dir`, `raw_data_path`) — the path BYOC customers have today for per-run / per-trigger overrides, pending a settable deployment-wide default.

## Context

Customer (Jan, on behalf of Prima Mente and Third Dimension) asked whether the metadata/raw-data bucket separation documented on this page is available for BYOC. The page didn't make that clear, and along the way Ketan caught that the section describing the metadata bucket's contents was just wrong.

Tracking issue: [DOC-1181](https://linear.app/unionai/issue/DOC-1181/check-metadata-separation-option-for-byoc-customers-in-helm-chart)

## Test plan

- [ ] `make dev` and visit `/deployment/selfmanaged/configuration/data-retention/` in the Union variant; verify both new callouts render and the anchor `#per-run-customization` resolves
- [ ] Confirm the relative link `../../../user-guide/task-deployment/run-context#storage` resolves (same depth as the existing `Traces` link on this page)
- [ ] Sanity-check the wording with Ketan before merging

### DIFF
--- a/content/deployment/selfmanaged/configuration/data-retention.md
+++ b/content/deployment/selfmanaged/configuration/data-retention.md
@@ -2,63 +2,72 @@
 title: Data retention policies
 weight: 7
 variants: -flyte +union
-description: Implications of object storage retention or lifecycle policies on the default bucket and metadata.
+description: Implications of object storage retention or lifecycle policies on the data plane bucket.
 ---
 
 # Data retention policies
 
-{{< key product_name >}} relies on object storage for both **metadata** and **raw data** (your data that is passing through the workflow). Bucket-level retention and lifecycle policies (such as S3 lifecycle rules) that affect the metadata store can cause execution failures, broken history, and data loss.
+The data plane uses an object store bucket (S3, GCS, or ABS) to hold the **raw data** of your workflows — files, directories, dataframes, models, and other large values offloaded from task inputs and outputs. Bucket-level retention and lifecycle policies (such as S3 lifecycle rules) on this bucket purge raw data and can break historical executions, caches, and trace-based recovery.
 
-## How {{< key product_name >}} uses the default bucket
+## Where metadata vs. raw data lives
 
-The platform uses a **default object store bucket** in the data plane for two distinct purposes:
+It helps to be precise about what "metadata" means here, because the term is used in two very different ways elsewhere in the industry.
 
-1. **Metadata store** — Task **inputs and outputs** passed between tasks at runtime, serialized as Protobuf literal messages. For primitive types (ints, strings, etc.) the literal carries the value directly; for offloaded types (`FlyteFile`, `FlyteDirectory`, `DataFrame`, etc.) it carries a pointer into the raw data store. The execution engine reads and writes these literals to pass results from one task to the next, and the UI reads them to render input/output values for both in-progress and historical runs.
+**Metadata in {{< key product_name >}} lives in the control plane database.** It includes:
 
-   > [!NOTE] What "metadata" does *not* mean here
-   > "Metadata" in this section refers strictly to data stored in the object store bucket — primarily task-to-task input/output literals and the {{< key product_name >}}-internal manifests that reference them. It does **not** refer to control plane database state (workflow and task registrations, execution status, schedules, audit history). That information lives in a separate relational database and is not affected by object-store retention policies.
+- Workflow, task, launch-plan, and artifact registrations.
+- Execution status, history, schedules, and audit trail.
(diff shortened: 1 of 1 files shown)


content/deployment/selfmanaged/configuration/data-retention.md
```diff
@@ -2,48 +2,76 @@
...
+## Where metadata vs. raw data lives
 
-The platform uses a **default object store bucket** in the data plane for two distinct purposes:
+It helps to be precise about what "metadata" means here, because the term is used in two very different ways elsewhere in the industry.
 
-1. **Metadata store** — References, execution state, and pointers to task outputs. The control plane and UI use this metadata to schedule workflows, resolve task dependencies, display execution history, and resolve output locations. This data is required for the correct operation of the platform.
+**Metadata in {{< key product_name >}} lives in the control plane database.** It includes:
 
-2. **Raw data store** — Large task inputs and outputs or complex types (for example `FlyteFile`, dataframes, etc.). The metadata store holds only pointers to these blobs; the actual bytes live in the raw data store.
+- Workflow, task, launch-plan, and artifact registrations.
```
fiedlerNr9: We dont have workflows, launch-plans or artifacts in V2. To give more exampls lets say:
- tasks
- triggers
- App definitions

content/deployment/selfmanaged/configuration/data-retention.md
```diff
@@ -2,48 +2,76 @@
...
 
-Data correctness is not silently violated, but the benefits of caching and trace-based recovery are lost for purged data.
+## Per-run customization
 
-## Customizing the metadata path
+Both the raw-data location and the engine's run base directory can be overridden **per run** via [`flyte.with_runcontext()`](../../../user-guide/task-deployment/run-context#storage):
 
-You can control where metadata is stored within the bucket via the **`config.core.propeller.metadata-prefix`** setting (e.g. `metadata/propeller` in the [data plane chart values](https://github.com/unionai/helm-charts/blob/master/charts/dataplane/values.yaml)). This lets you design lifecycle rules that **exclude** the metadata prefix (for example, in S3 lifecycle rules, apply expiration only to prefixes that do not include the metadata path) so that only non-metadata paths are subject to retention.
+- `raw_data_path` — storage prefix for offloaded raw data (`flyte.io.File`, `flyte.io.Dir`, `flyte.io.DataFrame`, checkpoints, etc.).
+- `run_base_dir` — base directory for the engine's per-run system data passed between tasks.
```
fiedlerNr9: I do not understand what `run_base_dir` is doing here regarding data retention. I would get rid of it

content/deployment/selfmanaged/configuration/data-retention.md
```diff
@@ -2,48 +2,76 @@
...
 
-Confirm the exact prefix and bucket layout for your deployment from the chart configuration, and validate any retention rules in a non-production environment before applying them broadly.
+This is the path BYOC customers have today for directing a run's raw data to a different bucket — for example, a customer-owned bucket with its own retention policy. Setting a deployment-wide default for these paths on BYOC is not currently supported.
```
fiedlerNr9: I would remove that. This page is for BYOK and I dont care about BYOC in here

fiedlerNr9 (approved):
Nice, that looks much better. Left some comments. 
Also, this is nice and generic now. Should we port it to BYOC as it is? 

I can also take note of a follow-up here, which is about explaining a backend config `max-cache-age` , which solves the issue of data retention on buckets for caching. (trying to load cached data which is already deleted)

PR #1089, opened 2026-06-06
selfhosted: document dataplane self-registration

mhotan:
## Summary

Extends `selfhosted/getting-started.md` Step 8 with a "Data plane self-registration" section explaining the operator's new self-registration flow:

- Lazy authz Resource creation on the first Heartbeat / UpdateStatus (no manual `provisionctl` or `uctl selfserve` step needed for selfhosted installs).
- The operator self-reports `connection_config` on every UpdateStatus so the control plane can dial back on the zero-trust SelectCluster path.
- The chart's `updateStatus.connectionConfig` block (`host`, `insecure`, `insecureSkipVerify`) drives the operator's self-report.

## Availability

Ships in the **2026.7.0** release. The section notes that self-registration requires **both the control plane and the data plane on 2026.7.0+** — reading the self-reported `connection_config` and constructing the dial-back target is added to the control plane in that release, so the control plane must be updated before (or with) the data planes that depend on it.

## Backward compatibility

The section explicitly notes that empty `host` keeps the legacy admin-set `DataplaneIngressURL` flow intact, so existing installs see no behavior change. The operator self-reports a **bare host**; the control plane constructs the `dns:///<host>:443` dial target and reverse-proxy URL from it (corrected from an earlier draft that said the chart formats the dial target).

## Related

Companion code changes ship in:
- Cloud PR: middleware Org-scope auth, lazy Resource creation, proto + cluster model column + SelectCluster fallback, operator self-reporting, terraform wiring.
- Helm-charts PR: chart values block + connection_config ConfigMap + operator mount.

### DIFF
--- a/content/deployment/selfhosted/getting-started.md
+++ b/content/deployment/selfhosted/getting-started.md
@@ -394,13 +394,13 @@ If authentication is enabled on the control plane, also set `AUTH_CLIENT_ID` in 
 ### Data plane self-registration
 
 > [!NOTE]
-> Self-registration is **opt-in as of 2026.7.0** and becomes the default in a future release; it requires the control plane and data planes on 2026.7.0 or later. Leaving it off preserves the legacy admin-set `DataplaneIngressURL` flow. See the helm-charts [control plane](https://github.com/unionai/helm-charts/releases/tag/controlplane-2026.7.0) and [data plane](https://github.com/unionai/helm-charts/releases/tag/dataplane-2026.7.0) release notes.
+> Self-registration is **opt-in as of 2026.7.0** and becomes the default in a future release; it requires the control plane and data planes on 2026.7.0 or later. Leaving it off keeps the existing behavior — the control plane routes to its statically-configured data plane URL. See the helm-charts [control plane](https://github.com/unionai/helm-charts/releases/tag/controlplane-2026.7.0) and [data plane](https://github.com/unionai/helm-charts/releases/tag/dataplane-2026.7.0) release notes.
 
 The data plane operator self-registers with the control plane on first contact — no manual provisioning step is required. On startup, the operator:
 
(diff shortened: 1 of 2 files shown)


content/deployment/selfhosted/getting-started.md
```diff
@@ -391,6 +391,32 @@ helm upgrade --install unionai-dataplane unionai/dataplane \
 
 If authentication is enabled on the control plane, also set `AUTH_CLIENT_ID` in your overrides file. See [Authentication](./authentication).
 
+### Data plane self-registration
+
+The data plane operator self-registers with the control plane on first contact — no manual provisioning step is required. On startup, the operator:
+
+1. Uses its existing OAuth client credentials (configured into the chart via `AUTH_CLIENT_ID` + secret) to authenticate to the control plane.
+2. The control plane's authorizer recognizes the identity (bound to the org admin policy at install time via the control plane chart's `services.authorizer.configMap.authorizer.bootstrap.serviceAccounts` block) and lazily creates the per-cluster authz Resource on the first `Heartbeat` and `UpdateStatus` call.
```
mhotan, line 402: Authorizer is only for Union authorizer type/mode.

content/deployment/selfhosted/getting-started.md
```diff
@@ -391,6 +391,32 @@ helm upgrade --install unionai-dataplane unionai/dataplane \
...
+3. On every `UpdateStatus` call, the operator reports a `connection_config` blob — the data the control plane needs to dial back to this data plane on the zero-trust execution path. This replaces the previous admin-set ingress URL flow.
```
mhotan: Too much internal Union jargon. All everyone needs to know is that the dataplane reports where to find itself to the control plane. The control plane either uses this endpoint directly or redirects clients in zero trust mode.

ppiegaze (approved):

PR #1095, opened 2026-06-09
Add Leases core-concepts page (union-only)

kumare3:
## What

Adds a new **Leases** page under User Guide → Core concepts (`+union -flyte`), plus a "Reliability" link-card to the Core concepts index (union variant only).

## Why

Users occasionally see `lease expired ...` messages and have no doc explaining what leases are or why they matter. This page is written for MLEs, DSs, and AI engineers — who understand **failures and cost**, not distributed systems — rather than platform internals.

## Contents

- **Rental analogy** — landlord/tenant, heartbeats as rent, grace period before eviction.
- **Why leases exist** — Union runs work across clusters it can't fully observe; a silent worker is indistinguishable from a dead one.
- **What `lease expired` means** — grace period, expiration, and failover; exactly-one-owner guarantee to prevent corruption.
- **The cost trade-off** — correctness over runaway compute: better to occasionally redo one action than pay for an unaccounted-for task.
- **Reassurance** — auto-heals momentary blips/deployments, expirations are rare, and what to actually do when you see one.

## Validation

- `make dist` has a **pre-existing** `four-o-four` layout panic on this checkout (reproduces identically on a clean tree with no changes — unrelated to this PR).
- Verified via the dev server (`make dev`): all 849 pages build with no errors. The new page renders (HTTP 200, title "Leases", `lease expired` content present, `{{< key product_name >}}` → Union.ai, cross-link to runs-and-actions resolves) and the Core concepts index renders the Leases card.

### DIFF
--- a/content/user-guide/core-concepts/_index.md
+++ b/content/user-guide/core-concepts/_index.md
@@ -30,3 +30,23 @@ When you run code with Flyte, here's what happens:
 4. Each task execution within a run is an **action**
 
 Let's explore each of these in detail.
+
+{{< variant union >}}
+
+{{< markdown >}}
+
+## Reliability
+
+How Union.ai keeps your work accountable when running across clusters.
+
+{{< /markdown >}}
+
+{{< grid >}}
+
+{{< link-card target="leases" icon="key" title="Leases" >}}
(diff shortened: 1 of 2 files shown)


davidmirror-ops (approved):
Thanks so much @kumare3 Really appreciate the analogies

PR #1123, opened 2026-06-22
Revert "Document interactive sandboxes (unionai-sandbox / union.sandbox)"

samhita-alla:
Reverts unionai/unionai-docs#1063


ppiegaze (approved):
Approving the interactive-sandboxes revert. Submodule bumped to infra/main 47c0ba1d so the redirects ship with the page removal (Check Redirects green); merging fires deploy-redirects.yml.

PR #1243, opened 2026-07-16
selfhosted auth docs: operator eager-key auto-provisioning

mhotan:
## Overview

Documents `config.operator.apiKey.enabled` as an automated alternative to the manual base64 `EAGER_API_KEY` secret step in the selfhosted auth guide: the data plane operator mints the key and writes it to the task-pod secret store (cluster-scoped, so multi-data-plane orgs don't share or rotate one credential). Notes it is currently opt-in (default `false`) and still needs the EAGER credentials resolvable on the control plane.

The `apiKeyOverrides` list interface landed separately on the base branch (#1264), so this PR was slimmed to only the operator-side automation that section doesn't cover.

## Test

Docs-only; prose + one YAML block appended to Step 5. Interface verified against the dataplane chart (`config.operator.apiKey.enabled`, `proxy.secretManager.enabled`).

### DIFF
--- a/content/deployment/selfhosted/authentication.md
+++ b/content/deployment/selfhosted/authentication.md
@@ -497,6 +497,22 @@ kubectl create secret generic <eager-secret-name> \
 > [!NOTE]
 > The exact secret name and key depend on your deployment's embedded K8s secret manager configuration. The secret name is typically an MD5 hash of a logical identifier. Contact {{< key product_name >}} support for the exact values for your organization.
 
+### Provision the task-pod key automatically with the operator
+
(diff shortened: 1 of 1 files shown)


content/deployment/selfhosted/authentication.md
```diff
@@ -462,25 +462,66 @@ kubectl create secret generic union-secret-auth \
...
+  --from-literal=client_id='<EAGER_CLIENT_ID>' \
+  --from-literal=client_secret='<EAGER_CLIENT_SECRET>' \
+  -n <controlplane-namespace>
+```
+
+```yaml
+services:
+  identity:
+    apiKeyOverrides:
+      EAGER_API_KEY:
```
mhotan: This format is incorrect now and needs to be updated.

PR #1253, opened 2026-07-21
Self Managed setup doc updates

davidmirror-ops:
Updating self managed setup docs so guidance is alligned with current product capabilities.

Changes include:

- Reducing `uctl` to the `selfserve` step
- Allign with the current "no generated values" approach
- Update some cloud-provider-specific instructions with recent findings from customer engagements
- Update with changes to EAGER API KEY process and queues

### DIFF
--- a/content/deployment/selfmanaged/selfmanaged-aws/deploy-dataplane.md
+++ b/content/deployment/selfmanaged/selfmanaged-aws/deploy-dataplane.md
@@ -9,7 +9,7 @@ variants: -flyte +union
 If you have not yet set up the required AWS resources (EKS cluster, S3, ECR, IAM), see [Prepare infrastructure](../selfmanaged-aws/prepare-infra) first.
 
 > [!NOTE] Planning more than one cluster?
-> This page covers the single-cluster path: one cluster in the `default` cluster pool, as created by the `flyte create cluster ... --pool default` command below. If you plan to connect several clusters to the same control plane, read [Multiple clusters](../configuration/multi-cluster) first. Pool membership governs metadata sharing -- clusters in the same pool share one metadata bucket, and clusters in different pools must use different ones -- so it affects the metadata bucket you configure below.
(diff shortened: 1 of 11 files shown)


content/deployment/selfmanaged/selfmanaged-aws/deploy-dataplane.md
```diff
@@ -30,59 +32,72 @@ If you have not yet set up the required AWS resources (EKS cluster, S3, ECR, IAM
    helm repo update
    ```
 
-2. Use the `uctl selfserve provision-dataplane-resources` command to generate a new client and client secret for communicating with your Union control plane, provision authorization permissions for the app to operate on the union cluster name you have selected, generate values file to install dataplane in your Kubernetes cluster and provide follow-up instructions:
+2. Configure the `flyte` CLI to talk to your control plane, then register the cluster name:
+
+   ```bash
+   flyte create config --endpoint <YOUR_UNION_CONTROL_PLANE_URL> --org <YOUR_ORG_NAME>
+   flyte create cluster <YOUR_SELECTED_CLUSTERNAME> --pool default
```
mhotan, line 42: I think `--pool default` is also optional. But we can leave to be explicit.

mhotan: We probably need to include some behavioral tradeoffs.

I think, we released changes recently to have the cluster-pool inherit the properties of the cluster upon heartbeat. This means that the cluster-pool is effectively pinned to that cluster's infrastructure until we can remove.

I still don't know if we can update or remove a cluster's cluster pool

mhotan (approved):

content/deployment/selfmanaged/selfmanaged-aws/deploy-dataplane.md
mhotan: I actually am not sure how to communicate all the limitations or pending work. Maybe leave out and see where things settle. Then be prepared for feedback.

davidmirror-ops: Yeah I think the idea is to cover the simplest path: single cluster on the default pool. I left a note pointing to multicluster docs (which also need to be updated).
The limitations are not blocking for most of our current customer's use cases

mhotan: Not being able to update a cluster's pool or delete a cluster seems like it will be limiting but I agree to just omit it and see what they say.

ppiegaze (approved):

PR #1259, opened 2026-07-22
docs: self-hosted multi–data-plane task routing and connectivity

mhotan:
**Context.** A self-hosted control plane fans work out to one or more data planes. Two things must be configured for that to work, and neither was covered end-to-end in the self-hosted docs: **which data plane runs a given project's tasks** (routing), and **how the control plane reaches each data plane** (connectivity). This PR documents both.

**What this PR does.**

*Task routing* — adds `operations/task-routing.md`, a new Operations guide covering the queue model (cluster pool → cluster → queue → `run.default_queue`) and the one-time steps to route each project to a data plane with the `flyte` CLI, with AWS and GCP `cluster-pool.yaml` examples and the control-plane-DB escape hatch for repointing the immutable cluster/queue → pool bindings. Adds a link card on the operations index.

*Connectivity* — clarifies the data plane's self-reported connection in `getting-started.md`. The self-registration section framed `updateStatus.connectionConfig` as a separate-cluster feature and implied a co-located data plane leaves `host` empty; both are wrong. States the general rule — the reported host **must be resolvable and network-reachable from the control plane**, since the CP dials it directly — gives the split (external DNS) vs shared-cluster (cluster-local Service address) guidance, adds the reverse-path `connectionConfig` block to the intracluster-topology section (mirroring the existing forward-path block), and fixes the empty-host note. Verified against a live intra-cluster deployment where the CP resolves the DP's self-reported `…svc.cluster.local` host and reverse-proxies to it.

**Design notes.** Routing config is create-only and additive (no rename/delete/repoint through the API yet). `union`-variant only. Docs-only.

Stacked on #932 (self-hosted deployment docs) — the `operations/` section and `getting-started.md` are introduced there. Retargets to `main` once #932 lands. (Supersedes #1260, folded in here.)

## Test plan

- [x] Hugo `union` variant builds cleanly (~1600 pages, no errors/warnings).
- [x] `operations/task-routing/` renders — title, `flyte create` commands, both AWS/GCP config tabs; `route` icon card on the operations index.
- [x] `getting-started` renders the reachability rule, the reverse-path block, and the cluster-local host example; the `#data-plane-self-registration` cross-anchor resolves.

### DIFF
--- a/content/deployment/selfhosted/operations/task-routing.md
+++ b/content/deployment/selfhosted/operations/task-routing.md
@@ -1,0 +1,165 @@
+---
+title: Task routing
+weight: 3
+variants: -flyte +union
+---
+
+# Task routing
+
+After installing a {{< key product_name >}} self-hosted deployment, the control plane still needs to be told **which data plane runs a given project's tasks**. Until that routing exists, a submitted run has no queue to land on, and control-plane bootstrap jobs that build images fail with `no enabled cluster in pool`.
+
+This guide covers the queue model and the one-time steps to configure routing for each data plane. Perform them after the control plane and data plane are installed and healthy, and after the data plane has registered with the control plane.
+
+## The queue model
+
+Routing a task requires three constructs, created in order:
+
+1. A **cluster pool** — carries the object store, secret store, and image registry contract that tasks in the pool use.
+2. A **cluster** subscribed to that pool. Subscribing a cluster implicitly creates a **queue** of the same name.
+3. A **`run.default_queue`** setting on each `(project, domain)` pair, pointing at a queue.
+
+A run submitted for a `(project, domain)` resolves to its default queue, which resolves to the pool the queue is bound to, which resolves to an enabled, healthy cluster in that pool.
+
+The simplest topology — and the recommended default — is **one pool, one cluster, one queue per data plane, all sharing the data plane's name.** You then point each project's `run.default_queue` at the data plane that should run its tasks.
+
+> [!NOTE]
+> Configuration is **create-only and additive.** At present there is no supported way to rename, delete, or repoint a cluster pool, cluster, or queue — draining is gated server-side. Choose data plane and pool names deliberately, and treat every step below as adding routing, never reconciling it. See [Changing existing routing](#changing-existing-routing) for the operational escape hatch when you must repoint an already-configured environment.
+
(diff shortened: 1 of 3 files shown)


content/deployment/selfhosted/operations/task-routing.md
```diff
@@ -0,0 +1,165 @@
...
+```
+
+{{< /markdown >}}
+{{< /tab >}}
+{{< /tabs >}}
+
+Create the pool:
+
+```shell
+flyte create cluster-pool my-data-plane --file cluster-pool.yaml
```
mhotan, line 79: we don't need to mislead users to think the pool name must match the cluster itself.

content/deployment/selfhosted/operations/task-routing.md
```diff
@@ -0,0 +1,165 @@
...
+You will hit this only on an environment that already accumulated a *different* pool topology — for example, a cluster or queue previously subscribed to a differently-named pool. The symptoms are:
+
+- On `flyte create cluster <name> --pool <name>`:
+  `cannot change cluster pool from "<old-pool>" to "<new-pool>"`.
+- On task routing or test runs:
+  `no enabled, healthy cluster in pool "<old-pool>"` — the queue still points at the old pool, which is now empty because the cluster moved.
+
+A fresh environment never hits this.
+
+### Repoint the bindings in the control-plane database
```
mhotan, line 119: We either need to omit this or clearly indicate this is dangerous operations with no guaranteed behavior for inflight runs.

content/deployment/selfhosted/operations/task-routing.md
mhotan: Clarified in d32dcb8a: the queue takes the cluster's name and the pool name is independent, so reusing the data plane's name across pool/cluster is now stated as a readability convention, not a requirement (queue model, Step 2, and the topology recommendation).

content/deployment/selfhosted/operations/task-routing.md
mhotan: Kept but gated behind a prominent WARNING in d32dcb8a: direct DB repointing is now flagged as unsupported and destructive, with no guaranteed behavior for in-flight runs (backup first, own-env only, prefer a fresh topology).

PR #1265, opened 2026-07-24
Adjust pools, clusters and queues docs

iaroslav-ciupin:


content/user-guide/cluster-workload-management/cluster-pools.md
```diff
@@ -19,19 +19,24 @@ secrets.
...
 ## Create a pool
 
-A pool is defined by its shared data plane contract: object store, secret store,
-and optional image registry.
+A pool's config is the data plane contract its member clusters must satisfy:
+object store, secret store, and image registry. The object store URI and secret
+store type are required here; the image registry is optional, and leaving it out
+means member clusters aren't checked against one. See
+[How a pool's config is established](#how-a-pools-config-is-established) for how
+that contract is enforced.
```
ppiegaze: `compareClusterPoolConfig` guards object store and secret store with `if expected != nil`, but compares image registry unconditionally via `GetLocator()` (nil -> `""`). So an unset pool registry requires clusters to report an *empty* one.

This asymmetry looks unintentional. If it's a code bug, fix the code and keep your original wording: that's the better outcome. This suggestion documents current behavior in the meantime.

```suggestion
A pool's config is the data plane contract its member clusters must satisfy:
object store, secret store, and image registry. The object store URI and secret
store type are required here. The image registry is optional to *set*, but it is
still compared: leave it unset and a member cluster that reports a registry is
marked unhealthy. See
[How a pool's config is established](#how-a-pools-config-is-established) for how
that contract is enforced.
```

content/user-guide/cluster-workload-management/cluster-pools.md
```diff
@@ -86,6 +91,44 @@ Supported `secret_store_type` values are `AWS_SECRETS_MANAGER`,
...
+   [Manage membership](#manage-membership)). The control plane accepts the
+   membership without inspecting the cluster's actual data plane.
+2. Once running, the cluster reports its real object store, secret store, and
+   image registry to the control plane in periodic status updates.
+3. The control plane compares each reported value against the pool's config. If
+   they don't match, the cluster is marked **unhealthy**. Only healthy, enabled
+   clusters are eligible for
+   [wildcard routing](./queues#how-a-queue-routes), so queues with the `*`
+   selector stop sending new work to that cluster until it recovers — the pool's
+   other healthy clusters absorb it.
```
ppiegaze: Two problems collide here, so I'm not proposing text until the blocking question is settled.

First: no comparison happens at all while the pool has at most one cluster (see the next comment). Second: the "so queues with the `*` selector stop sending new work" consequence is the claim I can't verify.

Once you confirm the routing behavior I'm happy to draft this step.

content/user-guide/cluster-workload-management/cluster-pools.md
```diff
@@ -86,6 +91,44 @@ Supported `secret_store_type` values are `AWS_SECRETS_MANAGER`,
...
+accepted first and only turns unhealthy a short time later. After registering a
+cluster into a custom pool, check `flyte get cluster <name>` to confirm it
+settles healthy rather than assuming the join succeeded.
+
+Two properties follow from this design:
+
+- **Image registry is optional.** Leave it unset in the pool config and the
+  control plane skips image registry validation entirely: whatever a member
+  cluster reports for its registry is accepted. Set it, and member clusters must
+  report a matching registry or go unhealthy.
```
ppiegaze:
```suggestion
- **Image registry is optional to set, but always compared.** Unlike the object
  store and secret store, which are skipped when the pool leaves them unset, the
  image registry is compared unconditionally against an empty value. Leaving it
  unset therefore requires member clusters to report *no* registry; a cluster
  that reports one goes unhealthy.
```

content/user-guide/cluster-workload-management/cluster-pools.md
```diff
@@ -86,6 +91,44 @@ Supported `secret_store_type` values are `AWS_SECRETS_MANAGER`,
...
+- **A pool with no config at all is populated by its first cluster.** The control
+  plane accepts a pool that declares nothing; the first cluster to join and report
+  its config then *populates* the pool config from what it reported, and every
+  cluster that joins later is validated against those values. The first reporter
+  wins, so if that cluster is misconfigured its values become the pool's contract
+  and correctly configured clusters are the ones that go unhealthy. Note that the
+  `flyte` CLI and `ClusterPool.create` both require an object store URI and a
+  secret store type, so creating a fully empty pool means calling the API
+  directly.
```
ppiegaze: `gormimpl/cluster.go:112-120`:

```go
mustPopulateConfig := assignedPool.JSONConfig == nil
if assignedPool.JSONConfig != nil {
    // ...count clusters...
    mustPopulateConfig = numClustersInPool <= 1
}
```

The trigger is "no config **or** at most one member cluster", so a pool created with explicit config is silently overwritten by its first cluster. That also makes the "confirm it settles healthy after registering" advice above hollow for the first cluster: it always will, because its own values became the contract.

```suggestion
- **A pool with one cluster or fewer takes its config from that cluster.** The
  control plane populates the pool config from the reporting cluster whenever the
  pool has no config *or* has at most one member cluster. An explicitly
  configured pool is therefore silently overwritten by its first cluster, not
  just an empty one. Validation against the stored config begins only once the
  pool holds two or more clusters. The first reporter wins, so if that cluster is
  misconfigured its values become the pool's contract and correctly configured
  clusters are the ones that go unhealthy.
```

content/user-guide/cluster-workload-management/cluster-pools.md
```diff
@@ -153,19 +236,28 @@ Cluster.create("prod-us-east-1", cluster_pool_name="prod")
...
+The `default` pool cannot be deleted. A custom pool can be deleted only when it
+contains **no clusters and no queues**; otherwise the request is rejected. Delete
+the member [clusters](./clusters#delete-a-cluster) first, then the pool.
+
+> [!NOTE] Queue deletion is not available yet
+> Because queues cannot currently be deleted, and registering a cluster creates a
+> [co-named queue](./clusters#the-co-named-queue) in that cluster's pool, a pool
+> that has ever had a cluster registered in it retains that queue and cannot be
+> deleted until queue deletion ships. In practice, today you can only delete a
+> custom pool that no cluster ever joined.
```
ppiegaze: Two reasons this is now wrong. `clusterRepo.Delete` deletes the co-named queue along with the cluster, so removing the clusters *does* free the pool. And `1683af0` shipped `DeleteQueue` in v0.6.0a2, so user-created queues can be removed too.

As written this tells people a custom pool is effectively undeletable, which will stop them doing something that works.

```suggestion
> [!NOTE] What counts as "no queues"
> Deleting a cluster also deletes its [co-named queue](./clusters#the-co-named-queue),
> so removing a pool's clusters clears the queues that came with them. Any queue
> you created in the pool yourself must be deleted separately with
> `flyte delete queue` before the pool can go.
```

content/user-guide/cluster-workload-management/clusters.md
```diff
@@ -115,39 +143,99 @@ which queues are bound to it, useful when deciding where to route or pin a queue
...
+or wait for its in-flight work to finish.)
 
 Deleting a cluster automatically removes it from the selector of every queue
 that pins it explicitly; wildcard (`*`) queues are unaffected. A queue whose
 selector becomes empty stops routing work anywhere until you point it at
-another cluster **in its pool**. A queue's pool can never change, so if the
-replacement cluster lives in a different pool, create a new queue there
+another cluster **in its pool**. You cannot move a queue to another pool, so if
+the replacement cluster lives in a different pool, create a new queue there
 instead.
```
ppiegaze: This is pre-existing text, but the code moved under it. `clusterRepo.Delete` calls `partitionExplicitQueues` and returns `FailedPrecondition` ("remove the cluster from those queues before deleting") for any explicit reference other than the co-named queue. Nothing is auto-unpinned, and the co-named queue is deleted rather than emptied.

```suggestion
Deleting a cluster is **rejected** while any queue other than its co-named queue
still names it explicitly: unpin the cluster from those queues first. Wildcard
(`*`) queues never block the delete. The cluster's
[co-named queue](#the-co-named-queue) is deleted along with the cluster.
```

content/user-guide/cluster-workload-management/clusters.md
```diff
@@ -56,20 +56,46 @@ Cluster.create("prod-us-east-1", cluster_pool_name="prod")
...
+`default` queue lives in the `default` pool with the `*` selector, so it routes
+to every healthy, enabled cluster in the `default` pool: a cluster registered
+there joins it automatically, while a cluster in any other pool never does.
+
+Both are ordinary queues — they appear in `flyte get queue`, carry the same
+concurrency, depth, priority, and fairness settings as any other, and are managed
+the same way on the [Managing queues](./queues) page. Two behaviors are specific
+to the co-named queue: it follows its cluster if the cluster is
+[reassigned to another pool](#move-a-cluster-to-a-different-pool), and its
+selector empties out if the cluster is [deleted](#delete-a-cluster).
```
ppiegaze: Same finding as the delete section below: the co-named queue is deleted with the cluster, not left with an empty selector.

```suggestion
to the co-named queue: it follows its cluster if the cluster is
[reassigned to another pool](#move-a-cluster-to-a-different-pool), and it is
deleted along with the cluster if the cluster is [deleted](#delete-a-cluster).
```

content/user-guide/cluster-workload-management/queues.md
```diff
@@ -59,19 +59,56 @@ flowchart TD
...
 
-> [!NOTE] Queue scope
-> Queues are currently organization-scoped. Some CLI and Python surfaces expose
-> `project` and `domain` parameters for future scoped queues, but current
-> deployments reject project/domain-scoped queue creation.
+There is one exception, and it isn't a queue update: a cluster's co-named queue
+follows that cluster if the cluster is
+[reassigned to another pool](./clusters#move-a-cluster-to-a-different-pool). Any
+other queue pinned to that cluster must be repointed first, or the reassignment
+is rejected.
```
ppiegaze: `8199989` (v0.6.0a2) makes queue pool-moves possible, so "the pool a queue lives in is **fixed at creation**" and "There is one exception" both need rewriting. From the docstring the move is gated (target pool must exist, the queue's clusters must be members of it, and the queue must be `drained`), so the isolation-boundary reasoning survives intact; it's the flat "never" that doesn't.

Leaving the rewrite to you since you know the intended framing. Same issue at `_index.md` ("a run or a queue can never move from one pool to another *in place*") and `clusters.md:237` ("You cannot move a queue to another pool").

content/user-guide/cluster-workload-management/clusters.md
```diff
@@ -115,39 +143,99 @@ which queues are bound to it, useful when deciding where to route or pin a queue
...
+### Before you move a cluster
+
+1. **Repoint custom-named queues.** Any queue *other* than the one named after
+   the cluster that pins it must be repointed or removed from the cluster first,
+   or the move is rejected. The cluster's [co-named queue](#the-co-named-queue)
+   moves to the destination pool with it, automatically.
+   `flyte get cluster <name>` lists the queues bound to the cluster.
+2. **Let runs finish.** [Draining](./queues#drain-and-reactivate-a-queue) is
+   coming in a future release; until it ships, wait for in-flight work to finish,
+   watching with `flyte get queue <name> --watch`.
```
ppiegaze: Small thing I couldn't resolve from the code: `Queue.drain()` and `flyte update queue --drain` have existed since v0.5.2, and the leasor has a full `Draining`/`Drained` state machine with a drain-completion loop. But the drain section says the control plane *rejects* drain requests, which is a server-side gate I can't see from the SDK.

Is that gate still in place? If it lifted, this item and the one at line 231 both change.

content/user-guide/cluster-workload-management/clusters.md
```diff
@@ -115,39 +143,99 @@ which queues are bound to it, useful when deciding where to route or pin a queue
...
+   executions, which {{< key product_name >}} still supports today. There is
+   currently **no way to see how many apps or v1 executions are running on a
+   given cluster**, and the queue precondition above does not account for them,
+   so nothing will block the move. If any are running when you reassign the
+   cluster, the cluster can be marked unhealthy and drop out of scheduling until
+   the mismatch is resolved. Confirm out-of-band that the cluster is idle before
+   moving it.
+4. **Make sure the configs match.** The destination pool's config must match
+   what the cluster reports, or the cluster goes unhealthy shortly after the
+   move — see below.
```
ppiegaze, line 229: Depends on the at-most-one-cluster finding above. When the destination pool has at most one cluster, the cluster doesn't go unhealthy: it silently rewrites the destination pool's config instead. That's arguably the more dangerous outcome and currently isn't documented anywhere. Worth calling out explicitly once the mechanism above is settled.

ppiegaze (commented):
Docs review from the Documentation side. I verified every claim below against `unionai/cloud@main` and `flyteplugins-union@main`; code refs are included so you can check my reading. One blocking question up top, then five accuracy items, then some staleness that's nobody's fault.

## Blocking question first, because four sections depend on the answer

The new NOTE says a `*` selector routes only to clusters that are **healthy and enabled**, and that an unhealthy cluster stops receiving new work until it recovers. I can't find that behavior. In `leasor/pkg/lease/scheduler.go`, `filterWorkersByClusters` returns **all** connected workers for `["*"]` with no health or state filter, and I found no read of the cluster `Health` field anywhere in the scheduling path (it's used for list filters, metrics, and display). A pool-config mismatch sets a DB health flag; it doesn't disconnect leaseworkers.

This matters because the mismatch -> unhealthy -> no-new-work chain appears in **four** places and is what tells the reader a misconfigured cluster fails safe. If it isn't true, we're reassuring people about something that will quietly keep taking work. Could you confirm which is right? Everything else here is cheap to fix; this one changes what several sections should say.

## Five things that don't match the control plane

Details inline.

1. Image registry is compared even when the pool leaves it unset (`compareClusterPoolConfig`, `gormimpl/cluster.go:562`), so "skips validation entirely" is backwards. Possibly a code asymmetry rather than a docs bug: see the inline note.
2. Pool config is overwritten whenever the pool has **at most one cluster**, not only when it's empty (`gormimpl/cluster.go:112-120`). An explicitly configured pool gets silently rewritten by its first cluster.
3. Deleting a cluster doesn't auto-unpin queues; it's **rejected** until you unpin, and the co-named queue is **deleted** with the cluster (`clusterRepo.Delete` -> `partitionExplicitQueues`).
4. Following from 3, the pool-deletion NOTE is inverted: deleting the clusters *does* free the pool.
5. Minor: reassigning a cluster to its current pool returns `AlreadyExists` (`cluster/service/cluster.go:925`), and "Manage membership" dropped the still-true fact that re-registration naming a different pool is rejected. Both paths exist now, so both are worth stating.

## Overtaken by your own work since this PR was opened

`flyteplugins-union` **v0.6.0a2** (2026-08-06) contains `8199989` (queue pool-moves) and `1683af0` (`DeleteQueue`/`UndeleteQueue`). Several "not yet available" statements are now false, including one in the drain section this PR doesn't touch. Flagged inline where I could anchor it.

## Release timing

`flyte update cluster --pool/--yes` and `Cluster.update()` are in **v0.6.0a2 only**, a prerelease; latest stable is `0.5.2`. Same for `cluster.pools` -> `cluster.pool` (correct from 0.6.0b0, wrong for 0.5.2). Our convention is to publish against a shipped release, so this content may want to wait or be split out. Your call on which.

## Two things I couldn't anchor to a diff line

- `prod-queue` now has contradictory selectors: `_index.md` shows `[Cluster C, Cluster D]` and teaches "membership is frozen"; `queues.md` still shows `selector: *` and says it "spreads across the eligible clusters". Same queue, same example topology, opposite lessons. Updating the `queues.md` mermaid block would resolve it.
- `queues.md` "Drain and reactivate a queue" still says "queues cannot be deleted", falsified by `1683af0`.

## Housekeeping

The branch is conflicted against `main` (`mergeable: false`, `rebaseable: false`): `clusters.md` collides with #1303, and main is 53 commits ahead. Also 13 em-dashes in the added lines; we're mid-sweep removing those corpus-wide (DOC-1257), so it would help not to reintroduce them.

To be clear about what's good here: the reassignment mechanics match `gormimpl/cluster.go` exactly (co-named queue migrates, other explicit refs block, nothing drains or reschedules, all atomic), the pool-deletion constraint matches `clusterPoolRepo.Delete` exactly, and the apps-and-v1-executions caveat is genuinely valuable operational knowledge that isn't derivable from the code. Naming the co-named queue as a concept and the "Before you move a cluster" checklist are real improvements. I'd like to keep all of it.

ppiegaze:
Heads-up on a merge-order change, so this does not get quietly clobbered later.

**#1270 (the selfmanaged restructure) is being sequenced ahead of this PR**, and it removes or moves all eight `deploy-dataplane.md` files edited here. Verified with `--find-renames` against its head:

| This PR edits | #1270 does |
|---|---|
| `selfmanaged-{aws,gcp,generic,oci}/deploy-dataplane.md` | **deletes** (content collapses into one shared page) |
| `selfmanaged-azure/deploy-dataplane.md` | renames → `deploy/_index.md` (R056) |
| `selfmanaged-{coreweave,crusoe,nebius}/deploy-dataplane.md` | renames → `deploy/{coreweave,crusoe,nebius}.md` (R097-098) |

When #1270 lands, this branch will show four `modify/delete` conflicts. **The tempting resolution — accept the delete — silently discards the change.** Flagging it now so nobody hits that cold.

**The re-application is not a copy-paste, and this is the part worth knowing.** I checked whether the paragraph this PR rewrites survives into #1270's structure. It does not:

- The text this PR replaces (*"Every organization is provisioned with a `default` pool that new clusters join automatically, so a single-cluster deployment needs no extra pool setup"*) is **absent from all four of #1270's targets.**
- #1270 rewrote that area, presumably absorbing #1261's registration step. What sits there now is: *"Register the cluster before you install the chart: the data plane binds to this record when it starts. Every organization is provisioned with a `default` pool, so `--pool default` needs no extra setup."* At `deploy/_index.md:59`, `deploy/coreweave.md:45`, `deploy/crusoe.md:45`, `deploy/nebius.md:47`.

So the sentence you are correcting has already been partly rewritten by someone else. What this PR adds that the new wording still lacks is the substantive part: **each cluster is assigned exactly one pool**, and **the default applies only when no pool is specified at creation**. That needs re-expressing against the new sentence rather than pasted over it.

Net effect: eight identical edits become four, but each one is a small rewrite, not a re-apply.

Nothing else in this PR collides. The other four files (`cluster-workload-management/*`) are untouched by #1270 except `_index.md`, where the two edits sit in different hunks and merge cleanly.

This is separate from the review feedback already on this PR, which is still open and unaddressed.

---
— docsy · automated docs agent

PR #1419, opened 2026-08-11
lance plugin docs

samhita-alla:


ppiegaze (commented):
Went through this against `flyte-sdk@origin/main`, the pinned examples commit, and by running the blob and handler-lookup paths locally. The plugin description is accurate throughout: the four handlers, the default-format split, the `storage_options` table including `aws_allow_http` and the `aws_endpoint` spelling, the streaming encoder, and the verbatim directory upload behind the reference handoff all match the source. All 18 code fragments resolve. Three things to fix.

One follow-up that isn't inline: `content/api-reference/integrations/` has `jsonl` and `polars` but no `lance`, and `api-packages.toml` is untouched. `26f6ed65` answered the 404 by dropping the link rather than generating the page. The package is on PyPI at 2.5.19, so adding the entry and regenerating would close it.

content/integrations/lance/_index.md
```diff
@@ -0,0 +1,272 @@
...
+
+{{< code file="/unionai-examples/v2/integrations/flyte-plugins/lance/multimodal_streaming.py" fragment="setup" lang="python" >}}
+
+Write it in chunks and hand it off by reference, which blob columns require anyway:
+
+{{< code file="/unionai-examples/v2/integrations/flyte-plugins/lance/multimodal_streaming.py" fragment="convert" lang="python" >}}
+
+### Streaming a shuffled epoch
+
+Draw a fresh random order each epoch, pull it in batches, straight out of object storage. Labels come from `take()` and the image bytes from `take_blobs()`, both scoped to the same rows:
```
ppiegaze, line 161: The example this introduces doesn't call `take_blobs()`. It pulls `image` through `take()`:

```python
batch = ds.take(order[i : i + batch_size], columns=["image", "label"])
for image, label in zip(batch.column("image").to_pylist(), ...):
    _ = len(image)  # stand-in for decoding and augmenting the image
```

That's exactly the trap the NOTE above warns about, in the one example people will copy into a real training loop. Ran it against pylance 10.0.0 with this schema:

```
take() schema:  image: struct<position: uint64, size: uint64>
values:         [{'position': 0, 'size': 8}, {'position': 64, 'size': 12}]
len(image):     2        # dict keys, not 8 bytes
take_blobs():   8, 12    # the actual bytes
```

So the loop never reads an image byte and `len(image)` is `2` for every row. "Peak memory is one batch" holds because the images aren't being fetched.

The fix is in the example rather than here: labels via `take(columns=["label"])`, bytes via `take_blobs("image", indices=...)`, which is what this sentence already describes. Needs a unionai-examples PR and a submodule re-bump.

content/integrations/lance/_index.md
```diff
@@ -0,0 +1,272 @@
...
+- **Point lookups:** Fetching individual records by id from a large table.
+- **Datasets bigger than memory:** Stream them rather than partition them by hand.
+
+Both are formats on `flyte.io.DataFrame`, so one task can produce Parquet and another Lance in the same workflow. What you cannot do is move a dataset cheaply between them. A `lance` input arrives as a `lance.LanceDataset` or a `pyarrow.Table`, and there is no third option, so this is a choice you make per dataset rather than per task. [Other dataframe types](#other-dataframe-types) covers what that rules out.
+
+### How much does it matter?
+
+The [benchmark below](#measuring-the-difference) pulls 1,000 random rows out of 100,000, each carrying a 512-byte payload. Parquet reads 53.6 MB to answer that. Lance reads 0.54 MB.
+
+The 100x gap in bytes read is the number to trust. It doesn't depend on caching and it grows with the dataset. The wall-clock speedup is much softer: about 10x on local disk, about 3.5x on the same benchmark against object storage, and it moves with your row sizes and batch size. Run it on your own data before quoting either figure.
```
ppiegaze: `compare()` measures `pa.Table.nbytes`, the in-memory size of the decoded result, not bytes read from storage:

```python
lance_bytes = lance_ds.take(indices, columns=columns).nbytes
parquet_bytes = (await parquet_df.open(pa.Table).all()).nbytes
```

Lance's `take()` also reads manifest and page metadata, and Parquet's figure only lands near its on-disk size because `rng.randbytes()` payloads are incompressible. "It doesn't depend on caching" is true for the wrong reason: `nbytes` isn't an I/O measurement at all.

The point survives a rewording. Flyte's Parquet decoder really does materialize the whole table to answer a 1,000-row request, and the report image already says "materialized". Suggest that word here, on line 47, and on line 232.

content/integrations/lance/_index.md
```diff
@@ -0,0 +1,272 @@
...
+| `pyarrow.Table`      | input     | `lance` (opt-in)  | Materializes the dataset in memory, honoring column subsetting             |
+
+`lance.LanceDataset` already defaults to the `lance` format, so you never have to annotate it. `pyarrow.Table` doesn't, and stays on Parquet unless you say otherwise. That is the one place the format shows up in your type signatures.
+
+### Other dataframe types
+
+Those four are the entire surface, which is worth knowing before you design a workflow around it. The Polars plugin can hand data to pandas and PySpark because all three speak Parquet. Lance has no such common currency here, so there is no lance-to-pandas or lance-to-Polars handler. Declare `pd.DataFrame` for a `lance` input and the task fails on the way in:
+
+```text
+TypeTransformerFailedError: Failed to find a handler for
```
ppiegaze: This is a plain `ValueError`. `dataframe.py:892` raises it and nothing wraps it. `TypeTransformerFailedError` is a `ValueError` subclass, but the type engine passes this one straight through. Message body is exact:

```
builtins.ValueError
Failed to find a handler for <class 'pandas.DataFrame'>, protocol [file], fmt ['lance']
```

```suggestion
ValueError: Failed to find a handler for
```

PR #1437, opened 2026-08-14
docs(spark): document pod templates for driver, executor, and task environment

pingsutw:
## Why

The Spark page lists `driver_pod` / `executor_pod` in the parameter table but never explains how pod templates actually work, and the environment-level `pod_template` isn't mentioned at all. The two paths behave differently in ways that fail silently:

- The env-level template requires a container named `primary` in the spec, otherwise serialization raises `No primary container defined in the pod spec`.
- `driver_pod` / `executor_pod` specs are passed to the Spark operator verbatim and **replace** the base template — a container whose name isn't `spark-kubernetes-driver` / `spark-kubernetes-executor` is quietly dropped by the operator's webhook.

## What

Adds a **Pod templates** section with both approaches, when to use each, and the naming/merge rules. Behavior verified against `flyteplugins/go/tasks/plugins/k8s/spark/spark.go` (`createSparkPodSpec`, `createDriverSpec`, `createExecutorSpec`), not inferred.

Docs-only change.

https://claude.ai/code/session_01WD41ZGHa41amZVxwbc3w79

### DIFF
--- a/content/integrations/spark/_index.md
+++ b/content/integrations/spark/_index.md
@@ -57,6 +57,63 @@ spark_env = flyte.TaskEnvironment(
 | `driver_pod` | `PodTemplate` | Pod template for the Spark driver pod |
 | `executor_pod` | `PodTemplate` | Pod template for the Spark executor pods |
 
+### Pod templates
+
+There are two places to customize the driver and executor pods:
+
+- **`TaskEnvironment(pod_template=...)`** — the task pod spec, used as the base pod template for *both* the driver and the executor pods.
+- **`Spark(driver_pod=...)` / `Spark(executor_pod=...)`** — role-specific specs that *replace* that base template for the driver or the executor.
+
+#### On the task environment
+
+Use this when the same customization (security context, tolerations, node selector, volumes, labels) should apply to driver and executor alike. The spec must contain the primary container — the container Flyte injects the task image and command into:
+
+```python
+from kubernetes.client import V1Container, V1PodSecurityContext, V1PodSpec
+
+spark_env = flyte.TaskEnvironment(
+    name="spark_env",
+    plugin_config=spark_config,
+    pod_template=flyte.PodTemplate(
+        labels={"team": "data-eng"},
+        pod_spec=V1PodSpec(
+            containers=[V1Container(name="primary")],  # required: the primary container
+            security_context=V1PodSecurityContext(run_as_user=1000),
+        ),
+    ),
+    image=image,
+)
+```
+
+Flyte renames the primary container to `spark-kubernetes-driver` / `spark-kubernetes-executor` before handing the template to the Spark operator, so the operator's webhook patches the right container.
+
+#### On the plugin config
+
+Use `driver_pod` / `executor_pod` when the two roles need to differ. These specs are passed through to the operator verbatim — they are *not* merged with the environment's pod template, and the container you want patched must be named `spark-kubernetes-driver` or `spark-kubernetes-executor`. A container under any other name is ignored by the operator's webhook.
+
+```python
+from kubernetes.client import V1Container, V1PodSpec, V1ResourceRequirements
+
+spark_config = Spark(
+    spark_conf={...},
+    executor_pod=flyte.PodTemplate(
+        primary_container_name="spark-kubernetes-executor",
+        pod_spec=V1PodSpec(
+            containers=[
+                V1Container(
+                    name="spark-kubernetes-executor",
+                    resources=V1ResourceRequirements(requests={"ephemeral-storage": "9Gi"}),
+                )
+            ],
+        ),
+    ),
+)
+```
+
+Because a role-specific spec replaces the base template rather than merging into it, anything you still need from the environment's `pod_template` has to be repeated in the `driver_pod` / `executor_pod` spec.
+
+Pod templates require a cluster whose Spark operator and `SparkApplication` CRD support them. On older clusters the template is dropped and only the legacy fields (cores, memory, and the other `spark_conf` settings) take effect.
+
 ### Accessing the Spark session
 
 Inside a Spark task, the `SparkSession` is available through the task context:
--- a/content/integrations/spark/_index.md
+++ b/content/integrations/spark/_index.md
@@ -115,9 +115,6 @@ spark_config = Spark(
 
 `driver_pod` / `executor_pod` replace the *pod template* only. Pod-level settings that Flyte sends to the Spark operator as explicit fields (affinity, tolerations, node selector, scheduler name, pod security context, DNS config, host network, labels, annotations, env, and image) are still derived from the task environment's pod template, and the operator applies them on top of whatever the role-specific spec sets. Use `driver_pod` / `executor_pod` for container-level differences between the two roles, such as resource requests, volume mounts, or sidecars, and keep pod-level customization on the environment's `pod_template`.
 
-> [!NOTE] Requires a `SparkApplication` CRD with pod template support
-> Pod templates require a cluster whose `SparkApplication` CRD supports `spec.driver.template`. On an older CRD, or where the platform has disabled the feature, the template is dropped: pod-level settings from the environment's `pod_template` still reach the driver and executor through the operator's own fields, but `driver_pod` / `executor_pod` do not apply at all. There is no error, the settings are silently absent.
-
 ### Accessing the Spark session
 
 Inside a Spark task, the `SparkSession` is available through the task context:
--- a/content/integrations/spark/_index.md
+++ b/content/integrations/spark/_index.md
@@ -113,7 +113,7 @@ spark_config = Spark(
 )
 ```
 
-`driver_pod` / `executor_pod` replace the *pod template* only. Pod-level settings that Flyte sends to the Spark operator as explicit fields (affinity, tolerations, node selector, scheduler name, pod security context, DNS config, host network, labels, annotations, env, and image) are still derived from the task environment's pod template, and the operator applies them on top of whatever the role-specific spec sets. Use `driver_pod` / `executor_pod` for container-level differences between the two roles, such as resource requests, volume mounts, or sidecars, and keep pod-level customization on the environment's `pod_template`.
+Because a role-specific spec replaces the base template rather than merging into it, anything you still need from the environment's `pod_template` has to be repeated in the `driver_pod` / `executor_pod` spec.
 
 ### Accessing the Spark session
 
--- a/content/integrations/spark/_index.md
+++ b/content/integrations/spark/_index.md
@@ -61,12 +61,14 @@ spark_env = flyte.TaskEnvironment(
 
 There are two places to customize the driver and executor pods:
 
-- **`TaskEnvironment(pod_template=...)`** — the task pod spec, used as the base pod template for *both* the driver and the executor pods.
-- **`Spark(driver_pod=...)` / `Spark(executor_pod=...)`** — role-specific specs that *replace* that base template for the driver or the executor.
+- **`TaskEnvironment(pod_template=...)`**: the task pod spec, used as the base pod template for *both* the driver and the executor pods.
+- **`Spark(driver_pod=...)` / `Spark(executor_pod=...)`**: role-specific specs that replace that base pod template for the driver or the executor.
+
+Both use `flyte.PodTemplate`. See [Pod templates](../../user-guide/tasks/task-configuration/pod-templates) for the general rules and for the `kubernetes` package the `V1*` types below come from.
 
 #### On the task environment
 
-Use this when the same customization (security context, tolerations, node selector, volumes, labels) should apply to driver and executor alike. The spec must contain the primary container — the container Flyte injects the task image and command into:
+Use this when the same customization (security context, tolerations, node selector, volumes, labels) should apply to driver and executor alike. The spec must contain the primary container, the container Flyte injects the task image and command into, whose name defaults to `primary`:
 
 ```python
 from kubernetes.client import V1Container, V1PodSecurityContext, V1PodSpec
@@ -85,19 +87,20 @@ spark_env = flyte.TaskEnvironment(
 )
 ```
 
-Flyte renames the primary container to `spark-kubernetes-driver` / `spark-kubernetes-executor` before handing the template to the Spark operator, so the operator's webhook patches the right container.
+Flyte renames the primary container to `spark-kubernetes-driver` / `spark-kubernetes-executor` before handing the template over, so Spark configures the right container.
 
 #### On the plugin config
 
-Use `driver_pod` / `executor_pod` when the two roles need to differ. These specs are passed through to the operator verbatim — they are *not* merged with the environment's pod template, and the container you want patched must be named `spark-kubernetes-driver` or `spark-kubernetes-executor`. A container under any other name is ignored by the operator's webhook.
+Use `driver_pod` / `executor_pod` when the two roles need to differ. These specs are passed through verbatim as the driver or executor pod template; they are *not* merged with the environment's pod template.
+
+Name the container `spark-kubernetes-driver` or `spark-kubernetes-executor`. Spark selects the driver and executor container from the template by that name; if no container matches, it falls back to the first container in the list and configures that one instead, so a template with a sidecar ahead of the real container silently runs the sidecar as the driver. The canonical names are also what the Spark operator and Flyte's log links use to find the container.
 
 ```python
 from kubernetes.client import V1Container, V1PodSpec, V1ResourceRequirements
 
 spark_config = Spark(
-    spark_conf={...},
+    spark_conf={"spark.executor.instances": "2"},
     executor_pod=flyte.PodTemplate(
-        primary_container_name="spark-kubernetes-executor",
         pod_spec=V1PodSpec(
             containers=[
                 V1Container(
@@ -110,7 +113,10 @@ spark_config = Spark(
 )
 ```
 
-Because a role-specific spec replaces the base template rather than merging into it, anything you still need from the environment's `pod_template` has to be repeated in the `driver_pod` / `executor_pod` spec.
+`driver_pod` / `executor_pod` replace the *pod template* only. Pod-level settings that Flyte sends to the Spark operator as explicit fields (affinity, tolerations, node selector, scheduler name, pod security context, DNS config, host network, labels, annotations, env, and image) are still derived from the task environment's pod template, and the operator applies them on top of whatever the role-specific spec sets. Use `driver_pod` / `executor_pod` for container-level differences between the two roles, such as resource requests, volume mounts, or sidecars, and keep pod-level customization on the environment's `pod_template`.
+
+> [!NOTE] Requires a `SparkApplication` CRD with pod template support
+> Pod templates require a cluster whose `SparkApplication` CRD supports `spec.driver.template`. On an older CRD, or where the platform has disabled the feature, the template is dropped: pod-level settings from the environment's `pod_template` still reach the driver and executor through the operator's own fields, but `driver_pod` / `executor_pod` do not apply at all. There is no error, the settings are silently absent.
 
 ### Accessing the Spark session
 
--- a/content/integrations/spark/_index.md
+++ b/content/integrations/spark/_index.md
@@ -68,7 +68,7 @@ Both use `flyte.PodTemplate`. See [Pod templates](../../user-guide/tasks/task-co
 
 #### On the task environment
 
-Use this when the same customization (security context, tolerations, node selector, volumes, labels) should apply to driver and executor alike. The spec must contain the primary container, the container Flyte injects the task image and command into, whose name defaults to `primary`:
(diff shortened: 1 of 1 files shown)


fiedlerNr9 (approved):

ppiegaze:
## Review — good PR, four fixes needed before merge

Reviewed at `6a2bdc010` against `flyteorg/flyte` `origin/main` @ `5cbd171`, `flyte-sdk` tag `v2.6.2`, and upstream `kubeflow/spark-operator` + `apache/spark`.

**First, credit where it's due, and a correction to my own initial suspicion.** I flagged `flyteplugins/go/tasks/plugins/k8s/spark/spark.go` as looking like a Flyte 1.x path. **It isn't** — it exists on both branches, `main`'s version imports `flyteorg/flyte/v2` and `spark-operator/v2`, and `podtemplate.go` exists *only* on `main`. You cited the right code.

**Verified correct:** every symbol and type resolves at released tag `v2.6.2` (`pod_template`, `PodTemplate(labels/pod_spec/primary_container_name)`, `Spark(driver_pod/executor_pod)`). The `No primary container defined in the pod spec` failure is real (`task_serde.py:402`) and — nicely — **correctly scoped** to the env-level path only; it would have been easy to over-generalize that to `driver_pod`/`executor_pod` and you didn't. The rename to `spark-kubernetes-driver`/`-executor` is real and sits in the right branch of the code. Example 1 constructs and serializes cleanly.

Two sentences are wrong, though, and both fail in the direction that costs a reader most.

---

### 1. "A container under any other name is ignored by the operator's webhook"

Wrong on both halves. It is **Apache Spark**, not the operator's webhook, and nothing is ignored — Spark **falls back to the first container**:

```scala
// apache/spark KubernetesUtils.scala:116-141
case _ => logWarning("specified container ... not found on pod template, " +
                     "falling back to taking the first container")
```

The operator's mutating webhook runs on the live pod and never reads the template; `buildDriverPodTemplate` just serializes it to YAML for `spark-submit`.

So with the single-container example below, the name genuinely doesn't matter. It matters when there are **sidecars** — Spark promotes whichever container is first and overwrites its image, command, resources and ports, so a sidecar ahead of the real container silently runs *as the driver*. The advice is right; the mechanism and the failure mode are both wrong, and "ignored" is the opposite of what happens.

**Suggested:**

> Name the container `spark-kubernetes-driver` or `spark-kubernetes-executor`. Spark selects the driver and executor container from the template by that name; if no container matches, it falls back to the first container in the list and configures that one instead, so a template with a sidecar ahead of the real container silently runs the sidecar as the driver. The canonical names are also what the Spark operator and Flyte's log links use to find the container.

### 2. "Anything you still need from the environment's `pod_template` has to be repeated"

`createSparkPodSpec` (`spark.go:161-179`) always populates the explicit `SparkPodSpec` fields from the **env-derived** pod spec, regardless of `customPodSpec` — affinity, tolerations, nodeSelector, podSecurityContext, DNS config, schedulerName, hostNetwork, labels, annotations, env, image. The code comment right below is explicit:

> `// the operator treats explicit fields as overrides of the template.`

So repeating those in `driver_pod` is not just unnecessary — **they'd be overridden**. `customPodSpec` replaces only `spec.{driver,executor}.template`, and even that isn't quite verbatim (`EnableServiceLinks` is back-filled).

**Suggested:**

> `driver_pod` / `executor_pod` replace the *pod template* only. Pod-level settings that Flyte sends as explicit Spark fields — affinity, tolerations, node selector, scheduler name, pod security context, DNS config, host network, labels, annotations, env, and image — are still derived from the task environment's pod template and are applied on top of whatever the role-specific spec sets. Use `driver_pod` / `executor_pod` for container-level differences between the two roles (resource requests, volume mounts, sidecars); keep pod-level customization on the environment's `pod_template`.

### 3. Example 2 raises `TypeError`

`spark_conf={...}` isn't a placeholder — `{...}` is a **set containing `Ellipsis`**:

```
>>> type({...})   ->  <class 'set'>  {Ellipsis}
>>> dict({...})   ->  TypeError: object is not iterable
```

`Spark.__post_init__` only replaces `None`, so it constructs fine and fails at deploy. It also re-binds `spark_config`, shadowing the working one defined earlier on the page. Minimal fix: `spark_conf={"spark.executor.instances": "2"}`, or reuse the page's existing dict.

### 4. The older-operator caveat should come back

`609dac846` removed it. Everything in this section is gated on `spark.go:188`:

```go
if GetSparkConfig().EnablePodTemplate && podTemplateSupported(ctx) {
```

`podTemplateSupported` reads the CRD schema once and **returns false on any failure** (missing RBAC, API error, timeout), cached for the process lifetime. This matters most for example 1, because the code comment at `spark.go:168-171` says a pre-2.x CRD silently drops `podSecurityContext` — which is what that example leads with. The page is `+flyte +union`, so it renders for OSS and BYOC clusters where the customer controls the operator version.

The original wording was slightly off (pod-level fields *do* still apply via Spark's own fields). **Suggested:**

> Pod templates require a cluster whose `SparkApplication` CRD supports `spec.driver.template`. On an older CRD, or where the platform has disabled the feature, the template is dropped: pod-level settings from the environment's `pod_template` still reach the driver and executor through Spark's own fields, but `driver_pod` / `executor_pod` do not apply at all. There is no error; the settings are silently absent.

If you know every supported cluster already carries a modern CRD, this can be a one-line note instead — happy to defer to you on that.

---

### Non-blocking

- **4 em-dashes** added to a page that had **zero**. House style bans them in docs prose (`style-spec.md` §4a). Three become `:` or `,`; the fourth is in a line being rewritten under #2 anyway.
- **`primary_container_name` is a no-op on this path.** Example 2 sets it, implying it does something; `createExecutorSpec` reads only `GetPodSpec()`. flyte#7822 dropped the override deliberately. Either remove it or say it's inert.
- **The `primary` default is never stated**, which is why the two examples look asymmetric. One clause fixes it: *"`primary_container_name` defaults to `primary`."*
- **No cross-link to `user-guide/tasks/task-configuration/pod-templates.md`**, which already covers the primary-container rule and the `pip install kubernetes` requirement both examples need and neither mentions.

### One note on the approval

Copilot reviewed `34870dd` — the commit **before** `609dac846` removed the caveat — so its summary carries no signal about the current state. It also left no findings, only a restatement.

None of this is carelessness: the behaviour you're documenting is **six days old** (flyte#7822 → #7850 → #7853 merged 2026-08-13, released `v2.0.42` on 08-14). The "replace, so repeat everything" framing was true of #7822 in isolation and became misleading once #7853 landed. Documenting this fast is genuinely valuable; it's just the window where a page is most likely to capture an intermediate state.

---
— docsy · automated docs agent

pingsutw:
Thanks, this was a genuinely useful review. All four blocking items are fixed in d5b23ff5, plus the non-blocking ones.

I re-verified each claim against `flyteorg/flyte` @ `5cbd171` rather than taking them on trust:

- **#2 confirmed.** `createSparkPodSpec` (`spark.go:161-179`) populates affinity, annotations, labels, env, image, `PodSecurityContext`, DNS config, tolerations, scheduler name, node selector, and host network from the env-derived `podSpec` before the `customPodSpec` branch is ever reached, and `customPodSpec` only replaces `spec.Template` (`:189-195`, with `EnableServiceLinks` back-filled). The "repeat everything" advice was actively harmful. Rewritten as suggested.
- **#3 confirmed.** `dict({...})` → `TypeError: cannot convert dictionary update sequence element #0 to a sequence`. Replaced with `spark_conf={"spark.executor.instances": "2"}`.
- **#4 confirmed.** `podTemplateSupported` caches via `sync.Once` and returns false on any failure (`podtemplate.go:27-33`), and the `spark.go:168-171` comment about pre-2.x CRDs dropping `podSecurityContext` is exactly the field example 1 leads with. Caveat restored with your corrected wording.
- **Non-blocking, all confirmed.** `primary_container_name` is inert on this path (only `executorPod.GetPodSpec()` is read, `spark.go:283`) so I removed it rather than documenting a no-op; added the `primary` default; cross-linked the pod templates guide for the `pip install kubernetes` requirement; em-dash count on the page is back to 0.

**One thing I'd flag on #1.** You're right about the actionable part, so I took your wording nearly verbatim. But the mechanism attribution isn't as clean as the review implies: the flyte code comment at `spark.go:201-202` says "The operator's mutating webhook patches the container it finds by name". Both are true in different places, and the template path is the one the docs describe, so I avoided naming a component for the fallback behavior:

> Spark selects the driver and executor container from the template by that name; if no container matches, it falls back to the first container in the list and configures that one instead …

I dropped "the canonical names are what the Spark operator … uses" down to a trailing aside for the same reason. If the flyte code comment is itself wrong, that's worth a separate issue against the plugin.

I could not construct example 2 locally to confirm (`cryptography` fails to build on this machine, unrelated), but the change only removes a kwarg and swaps a set literal for a dict, so it is strictly safer than what you already verified.

pingsutw:
address the comment

ppiegaze:
## Re-review at `cde7d23d7` — two fixed, six cleared, one still blocking

Re-verified against `flyteorg/flyte` `origin/main` @ `170ccab0a`.

### Fixed, and one of them improved on my suggestion

**#1 (the webhook/"ignored" error) — resolved well.** The mechanism is now right, and you added a claim I hadn't made:

> The canonical names are also what the Spark operator and Flyte's log links use to find the container.

**I verified that rather than taking it on trust, and it's correct** — `spark.go:488-490` builds `PodLogContext` with `PrimaryContainerName: "spark-kubernetes-executor"` and a matching `ContainerContext`. Good addition; it gives the reader a second, independent reason to use the canonical name.

**#3 (`spark_conf={...}` `TypeError`) — fixed.**

**All four non-blocking items — cleared.** Em-dashes back to **0**, `primary_container_name` dropped from example 2, the `primary` default now stated, and the cross-link to the pod-templates page added.

### #4 (older CRD) — your call, and I'll narrow the ask

You dropped it twice, so I'll take that as deliberate and stop pushing for the paragraph. One thing I'd still like, though, and it's one sentence: `spark.go:169-172` says `podSecurityContext` **is invisible to a pre-2.x CRD** and *"users who set this field … will see this setting getting dropped."* Example 1 leads with `V1PodSecurityContext(run_as_user=1000)`, so the page's first example is precisely the case that fails silently on an old CRD. Since the page is `+flyte +union` and OSS users pick their own operator version, a clause like *"on a pre-2.x SparkApplication CRD, `security_context` is silently dropped"* would cover it. If Union-supported clusters are all modern and you're happy for OSS readers to hit it, say so and I'll drop it entirely.

### #2 — still blocking, and the evidence is stronger than what I first wrote

The closing sentence is the problem:

> Because a role-specific spec replaces the base template rather than merging into it, anything you still need from the environment's `pod_template` has to be repeated in the `driver_pod` / `executor_pod` spec.

**Following that advice does not work.** `createSparkPodSpec` (`spark.go:160-178`) *always* populates these from the **environment's** pod spec, whether or not `driver_pod` is set:

```go
Affinity, Annotations, Labels, Env, Image,
PodSecurityContext, DNSConfig, Tolerations, SchedulerName, NodeSelector, HostNetwork
```

and the code comment immediately below states the resolution order:

```go
// The legacy fields above are always populated … Where the CRD accepts it, additionally
// pass a pod spec through as the pod template; the operator treats explicit fields as
// overrides of the template.
```

So for pod-level settings the instruction is wrong twice over: repeating them is **unnecessary** (they already arrive from the environment), and it is **ineffective** (the environment-derived explicit fields *override* the template, so the repeated values lose). A reader who sets different tolerations in `driver_pod` gets the environment's tolerations, with no error.

Not even the template path is a pure replacement — `EnableServiceLinks` is back-filled from the environment when a custom spec is present (`spark.go:191-193`).

`driver_pod` / `executor_pod` genuinely do replace the **template**, so the two sentences above it are fine. It's this last one that turns a true statement about templates into wrong advice about settings.

**Suggested replacement:**

> `driver_pod` / `executor_pod` replace the pod *template* only. Pod-level settings that Flyte sends as explicit Spark fields — affinity, tolerations, node selector, scheduler name, pod security context, DNS config, host network, labels, annotations, env and image — still come from the task environment's pod template, and take precedence over the same fields in a role-specific spec. Use `driver_pod` / `executor_pod` for container-level differences between the roles (resource requests, volume mounts, sidecars), and keep pod-level customization on the environment's `pod_template`.

### Also

The branch is **BEHIND** `main` and needs updating before merge.

Everything else is ready. Fix that one sentence and I'm happy to see this go in.

---
— docsy · automated docs agent

pingsutw:
Addressed in c29be05a0, and the branch is now up to date with `main` (2bbcfb52b, merge rather than rebase so the earlier review commits keep their SHAs).

**#2 — fixed, with the framing split rather than replaced.** Your evidence is right, and it matches what I found independently at `5cbd171`: `createSparkPodSpec` populates the eleven explicit fields from the env-derived pod spec before the `customPodSpec` branch is reached, `customPodSpec` only replaces `spec.Template`, and there is no merge anywhere else in the file. But "replace, so repeat it" is still the correct instruction for everything that only lives in the template: volumes, volume mounts, sidecars, init containers, image pull secrets, priority class. Your replacement dropped that half. So the sentence now does both jobs:

> Because a role-specific spec replaces the base template rather than merging into it, anything you still need from the environment's `pod_template`, such as volumes, volume mounts, or sidecars, has to be repeated in the `driver_pod` / `executor_pod` spec. The pod-level fields Flyte passes to the operator separately (affinity, tolerations, node selector, scheduler name, pod security context, DNS config, host network, labels, annotations, env, and image) still come from the environment and take precedence over the same fields in a role-specific spec.

I left out `EnableServiceLinks`. It is a real exception to "verbatim", but a reader who needs to know that is reading the plugin source anyway.

**#4 — dropping it, deliberately.** Taking you up on the offer to close this out. Noted that example 1 leads with the one field the pre-2.x CRD eats.

Thanks for both passes; #1 and #3 in particular were worth catching.

ppiegaze:
## Correction: the clause you took from me is wrong for 6 of the 11 fields. Please don't merge it as written.

You adopted my precedence sentence in good faith and I have to walk it back. After you conceded I stress-tested my own claim against the upstream operator instead of stopping, and **the half I was pushing hardest is the half that doesn't hold.** Your original instinct was closer to right than my correction was.

Your `c29be05a` sentence:

> The pod-level fields Flyte passes to the operator separately (affinity, tolerations, node selector, scheduler name, pod security context, DNS config, host network, labels, annotations, env, and image) still come from the environment **and take precedence over the same fields in a role-specific spec**.

### What I got right, and what I over-generalised

The **mechanism** is right — `createSparkPodSpec` fills the eleven explicit fields from the env-derived spec before the `customPodSpec` branch, and `customPodSpec` only becomes `spec.Template`. We agree there.

**"Take precedence" is not.** I read *"the operator treats explicit fields as overrides of the template"* out of the code comment at `spark.go:181-186` and trusted it as a statement about upstream behaviour. It is a Flyte-side comment about a third-party project, and per-field it is wrong four times. From `kubeflow/spark-operator` **v2.5.2** (the version `go.mod:98` pins), `internal/webhook/sparkpod_defaulter.go`:

| Field | Actual semantics | "takes precedence"? |
|---|---|---|
| `affinity` | `pod.Spec.Affinity = affinity.DeepCopy()` — wholesale | **yes** |
| `schedulerName`, `podSecurityContext`, `dnsConfig` | wholesale override when set | **yes** |
| `image` | overrides, but via Spark's `spark.kubernetes.driver.container.image`, not the webhook | yes, wrong attribution |
| **`tolerations`** | `pod.Spec.Tolerations = append(pod.Spec.Tolerations, tolerations...)` — **APPEND** | **no** |
| **`env`** | append to the container's env list | **no** |
| **`nodeSelector`** | per-key merge; keys only in the template **survive** | **no** |
| **`hostNetwork`** | one-way, only ever forces `true` | **no** |
| **`labels`, `annotations`** | see below — not from `pod_template` at all | **no** |

So a user who sets tolerations in `driver_pod` **keeps** them; they are unioned with the environment's, not replaced. The sentence tells them the opposite.

### `labels` and `annotations` are in the list by my error, and there may be a bug behind it

`spark.go:152-153` builds them from `TaskExecutionMetadata().GetLabels()/GetAnnotations()` unioned with `K8sPluginConfig` defaults — **not** from the environment's `pod_template`.

Worse, and worth checking before this page ships: `ToK8sPodSpec` returns the merged `ObjectMeta` as its second value, and `createDriverSpec`/`createExecutorSpec` **discard it** — `podSpec, _, primaryContainerName, err := ...` at `spark.go:225` and `:275`. The template is then built as `spec.Template = &v1.PodTemplateSpec{Spec: *templatePodSpec}` (`spark.go:211`) with **no `ObjectMeta` at all**.

Which reads as: **`TaskEnvironment(pod_template=flyte.PodTemplate(labels={...}))` never reaches the Spark pods by any route.** Example 1 on this page leads with `labels={"team": "data-eng"}`. If that's right the page is about to document something that silently does nothing — and I'd rather flag it as a question than assert it, since you know this plugin far better than I do. Worth a quick empirical check.

### Suggested

Keep your first sentence exactly as it is — it's correct and it's the half I wrongly talked you out of. For the second, either drop it, or scope it to what actually overrides:

> Pod-level scheduling and security settings that Flyte passes to the operator separately — affinity, scheduler name, pod security context and DNS config — come from the task environment and override the same fields in a role-specific spec. Tolerations, node selectors and environment variables are combined rather than replaced, so values set in `driver_pod` / `executor_pod` are kept alongside the environment's.

Or simply say nothing about precedence. The docs have never stated it, no example exercises it, and the honest position is that it's per-field and surprising.

### On process

I pushed this twice and was wrong on the substance both times; you were right to restore your wording, and right again to keep the "repeat it" half when you merged them. What I should have done before the first re-review is verify the operator's behaviour instead of quoting a comment about it. Sorry for the churn.

#1, #3 and the non-blocking items stand as fixed. Nothing else from me.

---
— docsy · automated docs agent

pingsutw:
Applied in 55854a6e1. Took the scoped version of your suggestion, with the em-dashes converted to parentheses for house style:

> Because a role-specific spec replaces the base template rather than merging into it, anything you still need from the environment's `pod_template`, such as volumes, volume mounts, or sidecars, has to be repeated in the `driver_pod` / `executor_pod` spec. Pod-level scheduling and security settings that Flyte passes to the operator separately (affinity, scheduler name, pod security context, and DNS config) come from the task environment and override the same fields in a role-specific spec. Tolerations, node selectors, and environment variables are combined rather than replaced, so values set in `driver_pod` / `executor_pod` are kept alongside the environment's.

**On the labels question: your read looks right, so I dropped it from the example.** `ToK8sPodSpec` returns the merged `ObjectMeta` as its second value (`pod_helper.go:816-832`), `createDriverSpec` and `createExecutorSpec` discard it (`spark.go:225`, `:275`), and `spec.Template` is built from the pod spec alone with no `ObjectMeta`. The `Labels` on the `SparkPodSpec` come from `TaskExecutionMetadata().GetLabels()` unioned with `K8sPluginConfig` defaults, which is a different source. There is no route left for `pod_template` labels to reach the driver or executor pods.

Rather than document a field that does nothing, example 1 now leads with `security_context` only, and `labels` is out of the "same customization" list above it. I have not filed anything against the plugin, since this is a static read and I could not run it. Worth an issue on `flyteorg/flyte` if someone can confirm empirically.

No complaint about the churn. Walking back your own claim after stress-testing it is the right call, and the page is more accurate for it.

ppiegaze:
## Good to merge from my side

`55854a6e1` resolves both. Re-checked the final text against the operator source rather than against my own earlier claim:

- **Overrides** — affinity, scheduler name, pod security context, DNS config. Correct: all four are wholesale replacements in `sparkpod_defaulter.go`.
- **Combined** — tolerations (`append`), node selectors (per-key merge), env (`append`). Correct.
- **Dropped** — `image`, `labels`, `annotations`, `host network`. Right call. `image` does override but via Spark's conf rather than the operator, and `hostNetwork` only ever forces `true`; leaving all four out is cleaner than caveating each.
- **`labels` out of Example 1** — good conservative call. The page no longer demonstrates something whose delivery path I couldn't find.

Branch is level with `main`, 14 green, only Build Determinism still running.

### One non-blocking nuance, for the record rather than for this PR

`addNodeSelectors` merges **per key**, and the environment wins on a collision:

```go
for k, v := range nodeSelector { pod.Spec.NodeSelector[k] = v }
```

So *"kept alongside the environment's"* holds for distinct keys but not for the same key. Narrow enough that I would not hold the PR for it, and adding the caveat may cost more clarity than it buys. Your call entirely.

### Worth a follow-up, separate from this page

Three things surfaced while verifying that are code-side, not docs-side:

1. **`primary_container_name` on `driver_pod`/`executor_pod` is silently ignored** since #7822 removed the override. The SDK still populates and sends it.
2. **The SDK's `PodTemplate` defaults its container name to `primary`**, while the plugin sets `podTemplateContainerName=spark-kubernetes-driver`. The obvious `Spark(driver_pod=flyte.PodTemplate(pod_spec=...))` may therefore name its container in a way Spark won't match. Flagged as a risk, not verified.
3. **The comment at `spark.go:242`** still says *"after the custom driver pod merge"*; there is no merge since #7822.

Happy to file these if useful, or leave them to you.

Thanks for the patience across three rounds — the two I was wrong about were the two I pushed hardest on.

---
— docsy · automated docs agent

PR #1438, opened 2026-08-14
docsy(v2): correct personnel-access and support-access claims in the security section

ppiegaze:
Corrects two of the three accuracy problems Ray Valencia raised in
[#team-docs](https://unionai.slack.com/archives/C01R4HFGR2T/p1786570525821129) on 2026-08-12, corroborated in
[#project-zero-trust](https://unionai.slack.com/archives/C0AQ0ULLKDL/p1786546808527469). Tracked as
[DOC-1424](https://linear.app/unionai/issue/DOC-1424).

Shalabh escalated this on 2026-08-13: *"Manas is concerned about our security posture and the team wants to
confirm the above before we point them to these docs."* These pages are live and are being cited into a
customer conversation right now.

### 1. `/security/` — the personnel-access overclaim

The index defines customer data to include workflow inputs and outputs, code bundles, logs and reports, then
asserts **"Union.ai personnel cannot access customer data or secrets."** Two separable claims are collapsed
there. Verified against `unionai/cloud@origin/main` (fetched 2026-08-14, `64ebb646`):

| Claim | Verdict | Evidence |
|---|---|---|
| Customer data does not transit the control plane | **True** | unchanged in this PR |
| Secret values cannot be read back | **True** | `idl/secret/proxy/definition.proto` — `Secret` carries `secret_id` + `SecretMetadata` only, no value field. `GetSecret` cannot return one. |
| Personnel cannot *view* customer data | **False as an absolute** | `shared_service/authorization/roles.go` — both `ROLE_TYPE_ADMIN` and `ROLE_TYPE_SUPPORT` carry `ACTION_VIEW_FLYTE_EXECUTIONS`. A holder of either sees run inputs and outputs, logs and reports in the console. |

The blurb now states the two true claims and describes the viewing surface accurately.

### 2. `/security/identity-and-access/human-access/` — support access framed as opt-in

Three sentences on this page are falsified by the shipped provisioning path:

- *"Customers must explicitly request it and configure the RBAC grants themselves"* and *"Union.ai staff
  cannot self-provision this access"* — `cli/uctl-admin/cmd/create/authorization.go` calls
  `GetUnionOktaUsers` at organization creation, which for a customer tenant resolves the **Union support
  staff** Okta group (`cli/uctl-admin/cmd/shared/authorization.go:532`, `shared_service/identity/utils.go`),
  and then assigns those users a policy via `AddAuthorizedOktaUsers`. Union provisions this; the customer
  neither requests nor configures it.
- *"This access can be time-limited so that it expires automatically after the support engagement
  concludes"* — `idl/authorizer/payload.proto` `AssignIdentityRequest` has fields `organization`, `identity`
  and a `role_id`/`policy_id` oneof. **There is no expiry, TTL or duration anywhere in the assignment API.**
  The two verification steps built on that time limit could never pass.

This PR removes the unsupportable absolutes and describes only the mechanism that is verified: an identity
in the tenant, bound to a policy, under the same authn, RBAC and audit logging as any other user. It
replaces the broken verification steps with `flyte get assignment` (documented in
`api-reference/flyte-cli.md`: "Without --user-subject, --creds-subject, or --email, lists all assignments")
followed by `flyte get policy <name>`.

**It deliberately does not assert the replacement policy.** "On by default, with the ability to opt out at
any time" is the reframing Ray floated, and neither he nor this PR can confirm a broad opt-out exists. See
the blocking questions below.

### 3. `/security/architecture/sovereign-data-plane/` — untouched, on purpose

Not in this PR. It needs a product and messaging decision, not a wording fix. Findings and options are in
[DOC-1424](https://linear.app/unionai/issue/DOC-1424).

## Review disposition and the open questions

EngHabu reviewed on 2026-08-19 and requested one change: drop the `no kubectl access` claim, because
opt-in support access to a cluster's Kubernetes API is being built and is not out yet. Applied in
`c3b9bdb`, in both places the page asserted it (the Self-managed lead and the self-managed
verification step). IAM roles, VPN and SSH keys are unchanged, and no other change was requested.

The claim was accurate at `unionai/cloud@fd0d5c53`: no shipped path grants Union personnel `kubectl`
against a self-managed customer data plane. The in-flight work is real and unmerged, in
`unionai/cloud` PRs [#17454](https://github.com/unionai/cloud/pull/17454) and
[#17607](https://github.com/unionai/cloud/pull/17607)
to [#17611](https://github.com/unionai/cloud/pull/17611): an in-cluster agent that fronts the API
server on a tailnet, off by default per cluster, with its own connection records. So the claim was
dropped for durability, not because it is false today.

The three questions that previously gated this draft are now tracked as their own tickets, so they no
longer hold the correction back. Each is about what the page could say **next**. Nothing in this PR
depends on their answers, and the text here asserts nothing they could falsify.

| | Question | Tracked at | State |
|---|---|---|---|
| 1 | What policy do Union support staff hold in a production customer tenant? `StaffPolicyName` (`cli/uctl-admin/cmd/shared/authorization.go:398`) returns admin outside staging. | [DOC-1487](https://linear.app/unionai/issue/DOC-1487/confirm-what-policy-union-support-staff-hold-in-a-production-customer) | Blocked (`eng-dep`) |
| 2 | Does a broad opt-out exist beyond the code viewer? | [DOC-1488](https://linear.app/unionai/issue/DOC-1488/confirm-whether-a-broad-customer-opt-out-from-union-support-staff) | Blocked (`eng-dep`) |
| 3 | Should the page disclose that the console hides support staff from the member list? | [DOC-1489](https://linear.app/unionai/issue/DOC-1489/decide-whether-the-security-docs-should-disclose-that-the-console) | Needs Scoping |

@EngHabu and @katrina: questions 1 and 2 are for you, on those tickets rather than here.

## Not changed, but flagged

`rbac.md` says Union.ai provides **three** user-assignable roles. `IsInternalSystemRole`
(`shared_service/authorization/roles.go`) hides only `cluster-manager`, `system-provisioned-access` and
`tenant-manager`, which leaves `support` and `flyte-project-admin-manager` user-visible. I did not edit the
role table because I could not verify what the console actually offers a customer admin.

---
— docsy · automated docs agent · [DOC-1424](https://linear.app/unionai/issue/DOC-1424)

### DIFF
--- a/content/security/identity-and-access/human-access.md
+++ b/content/security/identity-and-access/human-access.md
@@ -8,27 +8,25 @@ variants: -flyte +union
 
 ## Self-managed
 
-In self-managed deployments, Union.ai personnel access only the Union.ai-hosted control plane infrastructure. They have zero access to the customer's data plane. This access uses standard OIDC/SSO and RBAC.
+In self-managed deployments, Union.ai personnel have no infrastructure access to the customer's data plane: no IAM roles, no VPN, no SSH keys, and no `kubectl` access. They operate only the Union.ai-hosted control plane. Application-level access to the tenant is a separate matter, covered under [Customer-side support access](#customer-side-support-access) below.
 
 ## BYOC
 
 In BYOC deployments, Union.ai personnel additionally have authenticated Kubernetes cluster access for operational purposes: upgrades, node pool provisioning, Helm chart updates, health monitoring, and troubleshooting. This access uses cloud-native private connectivity (PrivateLink/PSC) and is scoped to Kubernetes cluster management. All cluster management actions are logged.
 
-## Customer-side support access (optional)
+## Customer-side support access
 
-Separately from BYOC Kubernetes cluster management, Union.ai offers an optional support service where customers can grant Union.ai staff access to the customer's tenant for troubleshooting. This is available for both self-managed and BYOC deployments.
+Separately from BYOC Kubernetes cluster management, Union.ai support staff hold an identity in the customer's tenant so that they can troubleshoot. This applies to both self-managed and BYOC deployments.
 
-When requested, Union.ai support personnel are granted access through the same RBAC framework used by the customer's own users. The customer creates a role binding for Union.ai staff, scoped to the specific projects, domains, and permission level appropriate for the troubleshooting engagement. This access can be time-limited so that it expires automatically after the support engagement concludes.
-
-This service is entirely optional. Customers must explicitly request it and configure the RBAC grants themselves. Union.ai staff cannot self-provision this access. The access is subject to the same authentication (OIDC/SSO), authorization (RBAC policies), and audit logging as any other user in the customer's organization.
+Support staff are granted access through the same RBAC framework that governs the customer's own users: an identity in the tenant, bound to a policy that determines what they can do. The access is subject to the same authentication (OIDC/SSO), authorization (RBAC policies), and audit logging as any other user in the customer's organization. Customers can list every identity-to-policy assignment in their organization, including any held by Union.ai staff, with `flyte get assignment`.
 
 This is distinct from BYOC Kubernetes cluster management access (described above), which is infrastructure-level access for platform operations. Customer-side support access operates at the application level: viewing runs, inspecting logs, diagnosing task failures, and reviewing configuration. It does not grant Kubernetes cluster access, IAM role access, or direct access to the customer's cloud account.
 
 ## Access scope
 
-When Union.ai personnel are granted access to a customer's tenant (in BYOC, or via the optional support service in self-managed), they *can*: view orchestration metadata, view logs relayed through the tunnel, perform administrative operations as authorized by the customer's RBAC policy, and (in BYOC) manage the Kubernetes cluster.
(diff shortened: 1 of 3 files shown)


content/security/identity-and-access/human-access.md
```diff
@@ -8,35 +8,33 @@ variants: -flyte +union
 
 ## Self-managed
 
-In self-managed deployments, Union.ai personnel access only the Union.ai-hosted control plane infrastructure. They have zero access to the customer's data plane. This access uses standard OIDC/SSO and RBAC.
+In self-managed deployments, Union.ai personnel have no infrastructure access to the customer's data plane: no IAM roles, no VPN, no SSH keys, and no `kubectl` access. They operate only the Union.ai-hosted control plane. Application-level access to the tenant is a separate matter, covered under [Customer-side support access](#customer-side-support-access) below.
```
EngHabu: I would drop the `no kubectl access` bit.. we are actively working on enabling opt-in support for that so users can request union support and audit this separately. It's not out yet

ppiegaze: Dropped, in both places the page asserted it: this line and the self-managed verification step at L41. IAM roles, VPN and SSH keys are unchanged. (`c3b9bdb`)

The lead clause still reads "no infrastructure access to the customer's data plane", which covers kubectl by implication once opt-in support access ships. Want that generalized too?

---
— docsy · automated docs agent · [DOC-1424](https://linear.app/unionai/issue/DOC-1424/security-section-states-intent-rather-than-current-behavior-personnel)

PR #1454, opened 2026-08-18
docsy(v2): Artifacts user-guide section

kumare3:
## Summary

Adds a new **Artifacts** section to the user guide, placed after **Apps** in the sidebar. The section is Union-only (`-flyte +union`).

An index page introduces artifacts (a named, versioned reference to an object in storage) with six cards, each backed by a full page:

- **Task outputs as artifacts** — `produces_artifacts=True` + `flyte.artifacts.new()`, metadata, model/data cards, and consuming artifacts as task inputs
- **Publish your own artifacts** — `Artifact.create()` and `flyte create artifact` for datasets and models produced outside tasks, with `external_ref` provenance
- **Prefetch Hugging Face models** — `flyte.prefetch.hf_model()`, versioned by HF commit, README attached as the model card, vLLM sharding
- **Trigger on new versions** — `flyte.OnArtifact` + `flyte.TriggeredArtifact`, fires regardless of who published the version
- **Use artifacts in apps** — `flyte.app.ArtifactValue` with deploy-time version pinning
- **Automatic lineage tracking** — producer and consumer edges, the Lineage tab in the UI, and the fetch vs tracked-consumption distinction

Sidebar weights of the sections after Apps are bumped by one to make room.

## Lineage view in the UI
lineage-ui

## Notes

- Content is grounded in the flyte-sdk implementation (`flyte.artifacts`, `flyte.remote.Artifact`, `flyte.prefetch`, `flyte.OnArtifact`, `flyte.app.ArtifactValue`) and the artifacts/triggers PRDs.
- `user-guide/tasks/task-configuration/triggers.md` still says artifact triggers are "coming"; worth updating in a follow-up to link here.

### DIFF
--- a/content/user-guide/_index.md
+++ b/content/user-guide/_index.md
@@ -71,6 +71,22 @@ Durable, self-healing agents built from tasks and apps, with sandboxing and MCP.
 
 {{< markdown >}}
 
+## Artifacts
+
+Name and version the datasets and models your tasks produce, and trace where they came from.
+
+{{< /markdown >}}
+
+{{< grid >}}
+
+{{< link-card target="artifacts" icon="box-seam" title="Artifacts" >}}
+Register task outputs and uploads as named, versioned artifacts, trigger runs on new versions, mount them into apps, and trace lineage.
+{{< /link-card >}}
+
+{{< /grid >}}
+
+{{< markdown >}}
+
 ## Access and identity
 
 How to authenticate and manage user permissions on your Union cluster.
--- a/content/user-guide/advanced-project/_index.md
+++ b/content/user-guide/advanced-project/_index.md
@@ -1,6 +1,6 @@
 ---
 title: Advanced project
-weight: 10
+weight: 11
 variants: +flyte +union
 mermaid: true
 llm_readable_bundle: true
--- a/content/user-guide/agents/_index.md
+++ b/content/user-guide/agents/_index.md
@@ -1,6 +1,6 @@
 ---
 title: Agents
-weight: 4
+weight: 5
 variants: +flyte +union
 ---
 
--- a/content/user-guide/apps/serve-and-deploy-apps/prefetching-models.md
+++ b/content/user-guide/apps/serve-and-deploy-apps/prefetching-models.md
@@ -103,7 +103,7 @@ Then run the CLI command:
 ```bash
 flyte prefetch hf-model meta-llama/Llama-2-70b-hf \
     --shard-config shard_config.yaml \
-    --accelerator L40s:8 \
+    --gpu L40s:8 \
     --hf-token-key HF_TOKEN
 ```
 
@@ -131,8 +131,10 @@ flyte prefetch hf-model <repo> \
     --hf-token-key HF_TOKEN \
     --cpu 4 \
     --mem 16Gi \
-    --ephemeral-storage 100Gi \
-    --accelerator L40s:4 \
+    --disk 100Gi \
+    --gpu L40s:4 \
+    --shm auto \
+    --raw-data-path s3://my-bucket/models \
     --shard-config shard_config.yaml
 ```
 
@@ -141,6 +143,20 @@ flyte prefetch hf-model <repo> \
 Here's a complete example of prefetching and using a model:
 
 {{< code file="/unionai-examples/v2/user-guide/serve-and-deploy-apps/prefetch_examples.py" fragment=complete-example lang=python >}}
+
+{{< variant union >}}
+{{< markdown >}}
+
+## Prefetched models are artifacts
+
+A successful prefetch registers the weights as a versioned model artifact, with the
+Hugging Face commit as its version and the repo README as its model card. For what
+that means for the registry, how to retrieve a prefetched model by name or by source
+repo, and how it appears in lineage, see
+[Prefetch Hugging Face models](../../artifacts/prefetch-models).
+
+{{< /markdown >}}
+{{< /variant >}}
 
 ## Best practices
 
--- a/content/user-guide/artifacts/_index.md
+++ b/content/user-guide/artifacts/_index.md
@@ -55,7 +55,7 @@ Union records who produced each artifact version and which tasks, triggers, and 
 The unit of work that produces and consumes artifacts.
 {{< /link-card >}}
 
-{{< link-card target="../apps" icon="rocket" title="Apps" >}}
+{{< link-card target="../apps" icon="window" title="Apps" >}}
 Long-running services that serve the models and datasets artifacts hold.
 {{< /link-card >}}
 
--- a/content/user-guide/artifacts/_index.md
+++ b/content/user-guide/artifacts/_index.md
@@ -1,0 +1,62 @@
+---
+title: Artifacts
+weight: 4
+variants: -flyte +union
+---
+
+# Artifacts
+
+An artifact is a named, versioned reference to an object in storage. Instead of remembering which bucket path holds last week's training set, you register the output under a name, and downstream tasks, triggers, and apps refer to it by that name.
+
+Your data path does not change. Files, directories, and dataframes are offloaded to object storage exactly as before. An artifact adds a name, a version, and a record of where it came from. Registering under the same name again creates a new version, and Union records which run produced each version and which tasks and apps consume it, so you can trace a deployed model back to the dataset it was trained on.
+
+Any `flyte.io.File`, `flyte.io.Dir`, or `flyte.io.DataFrame` can become an artifact. The most common way is to wrap a task's return value:
+
+```python
+@env.task(produces_artifacts=True)
+async def train() -> File:
+    weights = await File.from_local("model.pt")
+    return artifacts.new(weights, artifacts.Metadata(name="trained-model"))
+```
+
+{{< grid >}}
+
+{{< link-card target="task-outputs" icon="box-seam" title="Task outputs as artifacts" >}}
+Wrap a task's return value with `flyte.artifacts.new()` to register it as a named, versioned artifact with metadata and a model or data card.
+{{< /link-card >}}
+
+{{< link-card target="publish-artifacts" icon="cloud-upload" title="Publish your own artifacts" >}}
+Upload existing datasets and model weights from anywhere with `Artifact.create` or `flyte create artifact`, recording where they came from.
+{{< /link-card >}}
+
+{{< link-card target="prefetch-models" icon="download" title="Prefetch Hugging Face models" >}}
+Pull model weights from the Hugging Face Hub into your own storage as a model artifact, versioned by commit, with the model card attached.
+{{< /link-card >}}
+
+{{< link-card target="artifact-triggers" icon="lightning-charge" title="Trigger on new versions" >}}
+Run a task automatically whenever a new version of an artifact lands, using `flyte.OnArtifact`. It fires no matter who published the version.
+{{< /link-card >}}
+
+{{< link-card target="artifacts-in-apps" icon="window" title="Use artifacts in apps" >}}
+Mount model weights and datasets into serving apps with `flyte.app.ArtifactValue`, resolved and pinned at deploy time.
+{{< /link-card >}}
+
+{{< link-card target="lineage" icon="diagram-3" title="Automatic lineage tracking" >}}
+Union records who produced each artifact version and which tasks, triggers, and apps depend on it, and shows the graph in the UI.
+{{< /link-card >}}
+
+{{< /grid >}}
+
+## Related
+
+{{< grid >}}
+
+{{< link-card target="../tasks" icon="gear" title="Tasks" >}}
+The unit of work that produces and consumes artifacts.
+{{< /link-card >}}
+
+{{< link-card target="../apps" icon="rocket" title="Apps" >}}
+Long-running services that serve the models and datasets artifacts hold.
+{{< /link-card >}}
+
+{{< /grid >}}
--- a/content/user-guide/artifacts/artifact-triggers.md
+++ b/content/user-guide/artifacts/artifact-triggers.md
@@ -1,0 +1,52 @@
+---
+title: Trigger on new versions
+weight: 4
+variants: -flyte +union
+---
+
+# Trigger on new versions
+
+An artifact trigger runs a task automatically whenever a new version of a named artifact lands. Use it to validate every new model, retrain when a dataset is refreshed, or kick off batch inference when fresh data arrives.
+
+A trigger is a `flyte.Trigger` whose `automation` is `flyte.OnArtifact`. The `flyte.TriggeredArtifact` sentinel marks which task input receives the new version:
+
+```python
+import flyte
+from flyte.io import File
+
+env = flyte.TaskEnvironment(name="validation")
+
+retrain = flyte.Trigger(
+    name="retrain-on-new-model",
+    automation=flyte.OnArtifact(name="customer_model"),
+    inputs={"model": flyte.TriggeredArtifact, "threshold": 0.5},
+    description="Validate every new customer_model version",
+)
+
+
+@env.task(triggers=(retrain,))
+async def validate(model: File, threshold: float = 0.5) -> str:
+    ...
+```
+
+Deploy the environment to register the trigger:
+
+```bash
+flyte deploy validation.py env
+```
+
+From then on, every new version of `customer_model` in the task's project and domain starts a run of `validate`, with the new version bound to the `model` input. The task body sees an ordinary `File` or `Dir` and needs no artifact-specific code.
+
+## What counts as a new version
+
+The trigger fires no matter how the version was published: a task output wrapped with `flyte.artifacts.new()`, an upload through `flyte.remote.Artifact.create()` or `flyte create artifact`, or a Hugging Face prefetch. This makes triggers a clean handoff point between external processes and your pipelines: a partner drops a dataset, publishes it as an artifact, and your processing task starts on its own.
+
+By default any new version fires the trigger. Pass `version=` to `flyte.OnArtifact` to fire only when that exact version is published.
+
+## Rules
+
+* At most one input can be `flyte.TriggeredArtifact`.
+* `flyte.TriggeredArtifact` requires the automation to be `flyte.OnArtifact`.
+* `flyte.TriggerTime` cannot be combined with `flyte.OnArtifact`; it is for schedule triggers.
+
+All other inputs must have values in the trigger definition or defaults on the task, the same as [schedule triggers](../tasks/task-configuration/triggers).
--- a/content/user-guide/artifacts/artifacts-in-apps.md
+++ b/content/user-guide/artifacts/artifacts-in-apps.md
@@ -1,0 +1,42 @@
+---
+title: Use artifacts in apps
+weight: 5
+variants: -flyte +union
+---
+
+# Use artifacts in apps
+
+A serving app usually needs model weights or reference data on disk before it starts. Instead of baking them into the image or downloading them in startup code, declare an artifact as an app parameter and Union mounts it for you.
+
+`flyte.app.ArtifactValue` names the artifact, and the parameter's `mount` says where its contents appear in the container:
+
+```python
+import flyte
+import flyte.app
+
+env = flyte.AppEnvironment(
+    name="classifier-api",
+    resources=flyte.Resources(cpu="2", memory="8Gi"),
+    parameters=[
+        flyte.app.Parameter(
+            name="model",
+            value=flyte.app.ArtifactValue(name="trained-model", type="directory"),
+            mount="/tmp/model",
+        )
+    ],
+)
+```
+
+The app code reads `/tmp/model` like any local directory. Only `flyte.io.File` and `flyte.io.Dir` artifacts can be app parameters; set `type` to `"file"` or `"directory"` to match.
+
+The prebuilt serving environments accept an `ArtifactValue` directly. For example, the vLLM and SGLang app environments take one as `model_path`, which pairs naturally with [prefetched Hugging Face models](prefetch-models).
+
+## Versions are pinned at deploy time
+
+The artifact is resolved when the app is deployed. If you leave out `version`, the latest version at deploy time is chosen and pinned, so a new version published later does not swap weights under a running app. Redeploy to move to the newer version, or pass an explicit `version=` to stay on a known one.
+
+Union records exactly which artifact version each app deployment resolved, so the artifact's [lineage view](lineage) shows every app serving it.
+
+## Apps can publish artifacts too
+
+An app endpoint can call `flyte.remote.Artifact.create()` to register a new version from inside the app, for example to snapshot state it has built up. That version behaves like any other: it shows up in the registry and can fire [artifact triggers](artifact-triggers).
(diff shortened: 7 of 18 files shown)


ppiegaze:
## docsy review — Artifacts user-guide section

Assessed against the shipped source (`flyte-sdk` `origin/main` @ `5b8c0380` + released tags `v2.6.0`/`v2.6.1`, `flyteorg/flyte` @ `77e45d5a`, `unionai/cloud` @ `cb52d586`) and the built site. **44 claims checked; 36 CONFIRMED.** Reviewed at head `33846f70` after the force-push, so this applies to the finalized diff.

**Verdict: substantively good, not yet mergeable.** Seven blocking items, three of which ship code that cannot run.

First, credit where it's due, because this is unusual: **zero em-dashes, zero AI tell-vocabulary, sentence-case headings throughout, autolinker convention respected, weights 1–12 contiguous with no collisions** — including the two loose top-level pages (`authenticating.md`, `user-management.md`) that are the easy ones to miss. The page shapes mirror `apps/_index.md` correctly. All 14 CI checks pass. The structure is right.

---

## Blocking — code that does not run

**1. `flyte.AppEnvironment` does not exist** — `artifacts-in-apps.md`

Runtime-verified: `hasattr(flyte, 'AppEnvironment')` is `False`. The page's only example raises `AttributeError`.

```python
env = flyte.app.AppEnvironment(   # was: flyte.AppEnvironment(
```

Not your bug originally: `content/user-guide/apps/_index.md:14` already carries the same error. Filing that separately.

**2. `import flyte` does not expose `flyte.prefetch`** — `prefetch-models.md`

Runtime: `AttributeError: module 'flyte' has no attribute 'prefetch'`. Subpackages need an explicit import:

```python
import flyte
import flyte.prefetch          # add this
```

The other pages get this right (`from flyte import artifacts`, `from flyte.remote import Artifact`, `import flyte.app`).

**3. The sharding example cannot run** — `prefetch-models.md`

`hf_model` defaults to `Resources(cpu="2", memory="8Gi", disk="50Gi")` — no accelerator — but the example asks vLLM for `tensor_parallel_size=8` on a 70B model.

```python
run = flyte.prefetch.hf_model(
    repo="meta-llama/Llama-2-70b-hf",
    shard_config=ShardConfig(engine="vllm", args=VLLMShardArgs(tensor_parallel_size=8)),
    resources=flyte.Resources(cpu="8", memory="64Gi", disk="500Gi", gpu="A100:8"),
    hf_token_key="HF_TOKEN",
)
```

Heads-up: the SDK's own docstring for this passes `accelerator="A100:8"`, which is **not a parameter of `hf_model`**. That's an upstream docstring bug — don't copy it.

## Blocking — a false mechanism

**4. The WARNING box in `task-outputs.md` is right about the outcome, wrong about the cause.**

It says that without `produces_artifacts=True` the wrapper "is silently unwrapped and the task returns a plain value". Unwrapping and the `ProducedArtifact` declaration happen **regardless** of the flag — `convert.py:452`, and `tests/flyte/test_produces_artifacts.py:187` states it outright: *"Declarations do not depend on `produces_artifacts` — the flag only gates backend extraction."* The real gate is backend-side, at `cloud:leaseworker/action_handler.go:259-263`.

Suggested:

> **Both parts are required**
> The wrapper is stripped from the output value either way. Without `produces_artifacts=True` the platform never extracts the metadata, so no artifact is registered, no trigger fires, and there is nothing for an app to bind to. Nothing warns you.

## Blocking — corpus consistency

**5. `prefetch-models.md` duplicates an existing page, with the opposite variant gate.**

`content/user-guide/apps/serve-and-deploy-apps/prefetching-models.md` (173 lines, `variants: +flyte +union`) already covers HF prefetch in more depth — Python API, CLI, `--wait`, custom artifact name, HF token, sharding, and using prefetched models in vLLM/SGLang apps. This PR adds a second, shorter treatment gated `-flyte +union`.

They also contradict each other on flags: existing page uses `--ephemeral-storage` / `--accelerator`; this PR uses `--disk` / `--gpu`. Readers will hit whichever is wrong.

Cheapest fix: drop `artifacts/prefetch-models.md` and point the `_index.md` card at the existing page.

**6. `tasks/task-configuration/triggers.md` is left contradicting this PR.**

Lines 15–21 still read *"Currently, only **schedule triggers** are supported. Support is coming for other trigger types, such as: … Artifact triggers."* The PR body defers this, but merging as-is makes the corpus self-contradicting on the exact feature being added.

It has to be variant-guarded — that page is `+flyte +union`, these pages are `-flyte +union`, so a flat rewrite both asserts a Union-only capability to OSS readers and creates a `+flyte` → `-flyte` link that **fails `make check-links`**:

```
{{< variant union >}}
{{< markdown >}}
Two trigger types are available: **schedule triggers**, which run a task on a Cron
expression or fixed rate, and **artifact triggers**, which run a task when a new
version of a named artifact lands. See
[Trigger on new versions](../../artifacts/artifact-triggers).
{{< /markdown >}}
{{< /variant >}}
{{< variant flyte >}}
{{< markdown >}}
Currently, only **schedule triggers** are supported.
This type of trigger runs a task based on a Cron expression or a fixed-rate schedule.

Support is coming for other trigger types, such as webhook triggers: hit an API
endpoint to run your task.
{{< /markdown >}}
{{< /variant >}}
```

**7. The section is orphaned from the User guide landing page.**

`content/user-guide/_index.md` carries a link-card for all 11 other sections; Artifacts gets none. Confirmed on the preview — it's the only section with no card. **CI cannot catch this**: a *missing* link isn't a *broken* link, which is why everything is green.

The card must go **inside** the existing `{{< variant union >}}` block — the unguarded "Basics" grid renders in the `flyte` variant too, and a card there pointing at a `-flyte` page is a variant-safety bug.

---

## The variant gate — worth settling before anything else

This is the one editorial call I'd push back on, and the evidence cuts against `-flyte`:

- **Symbols: fully open source, already released.** Every symbol these pages document ships in the Apache-2.0 `flyte` SDK at **v2.6.0 and v2.6.1** — verified by `git ls-tree` at both tags plus live runtime import. Nothing is Union-only at the SDK layer.
- **Backend: Union-only today.** The artifact service exists solely in `unionai/cloud`. OSS `flyteorg/flyte` ships the **IDL only** (`git grep ArtifactService -- ':!gen' ':!flyteidl2'` → zero hits), no `charts/` deployment.
- **But the site already publishes these symbols to Flyte OSS readers.** All gated `+flyte +union` on `main` right now: `flyte.artifacts/*`, `flyte.remote/artifact`, `flyte.prefetch/*`, `flyte/onartifact`, `flyte.app/artifactvalue`, `flyte create artifact` and `flyte get artifact` in `flyte-cli.md`, `prefetching-models.md`, and `triggers.md` — whose trigger service is **equally** Union-only and which is `+flyte`. Live: `/docs/v2/flyte/api-reference/flyte-sdk/flyte.artifacts/` returns 200.

As submitted, the OSS variant ships a complete artifacts API + CLI reference with **zero** conceptual content explaining it, plus a `+flyte` prefetch page discussing artifacts while the artifacts section is invisible.

**Suggested: `variants: +flyte +union` on all seven pages**, with a `{{< variant union >}}` "requires the Union artifact service" callout on the backend-dependent behavior — which is exactly how `triggers.md` already handles the same dependency, and matches the Artifacts PRD's declared variants (Flyte OSS opt-in + BYOC + Self-managed).

If you'd rather keep `-flyte +union`, that's a legitimate call, but then the same PR needs to re-gate the 10+ reference pages above and reconcile `prefetching-models.md` — otherwise the split ships.

**One question no source can answer: is an artifact service planned for the OSS Flyte 2 backend?** The IDL and generated stubs are already in OSS with nothing behind them, which reads like a placeholder. That's an eng call, and it decides this.

---

## Non-blocking

- `lineage.md:30` — `birds-eye` should be `bird's-eye`; the sentence reads better deleted.
- `prefetch-models.md:33` — `commit id` → `commit ID` (corpus convention).
- `lineage.md` / `_index.md` — the tab is labelled **`Artifact card`** (sentence case) in `clientsv2`, not `Artifact Card`; the list also omits the **Overview** tab.
- `lineage.md` — "Each node links to its own page" is imprecise: trigger rows are *expandable* rather than navigations (`graph.ts:52-56`).
- `prefetch-models.md` — "the repo's README is attached as the model card" is true only **if the repo has one** (`_hf_model.py:575-596`).
- `prefetch-models.md` — "the CLI accepts the same options as the Python API" isn't quite right; the CLI splits resources into `--cpu/--mem/--disk/--gpu/--shm` and folds them into one `Resources(...)`. There is no `gpu=` Python kwarg.
- `artifacts-in-apps.md` — "an app endpoint can call `Artifact.create()`" is plausible but I found no test, example, or docstring asserting it, and `_current_task_source()` returns `None` outside a task, so such a version carries no provenance unless `external_ref` is passed. Worth verifying or dropping.
- `_index.md` — the Apps "Related" card uses `icon="rocket"`; the house convention elsewhere is `icon="window"`.
- **`flyte.TriggeredArtifact` and `flyte.TriggerTime` are in no linkmap**, so they won't autolink. Not a build break, but it suggests they're missing from the generated API reference — if they're public, that's a regen, not a docs edit.
- **Examples are inline rather than pulled from `unionai-examples`.** Three of these snippets shipped broken, which is precisely what the `{{< code file=… fragment=… >}}` convention prevents. `flyte-sdk/examples/artifacts/artifact_example.py` and `examples/triggers/artifact_trigger.py` are ready-made sources.

---

Happy to take any of this on directly if that's easier than round-tripping — say the word and I'll push the fixes to your branch or open a follow-up.

---
— docsy · automated docs agent · [DOC-1207](https://linear.app/unionai/issue/DOC-1207/docs-coverage-artifacts) · [run report](https://github.com/unionai/docsy/blob/main/runs/2026-08-18-docsy-artifacts-pr-1454.md)

ppiegaze:
## Fixes pushed to this branch — `570447c`

Peeter reviewed the findings and directed these. Every one of them disposed of explicitly:

| # | Finding | Disposition |
|---|---|---|
| 1 | `flyte.AppEnvironment` does not exist | **Fixed** → `flyte.app.AppEnvironment` |
| 2 | `import flyte` does not expose `flyte.prefetch` | **Fixed** → explicit subpackage import |
| 3 | Sharding example has no accelerator | **Fixed** → explicit `resources` + a sentence saying why |
| 4 | WARNING box states a false mechanism | **Fixed** → reworded; conclusion kept, mechanism corrected |
| 5 | `prefetch-models.md` duplicates `prefetching-models.md` | **Deliberately left** — Peeter's call, for now |
| 6 | `triggers.md` says artifact triggers are "coming" | **Fixed** → variant-guarded |
| 7 | Section orphaned from the User guide landing page | **Fixed** → card added inside the union-only block |

## Variant gating: `-flyte +union` confirmed as intended

I argued for `+flyte +union` on the grounds that the symbols are all in the OSS SDK and the site already publishes their API reference to Flyte OSS readers. **Peeter has confirmed `-flyte +union` is correct, so that is settled** and I have built to it. Fix 6 in particular depends on it: the flyte variant of `triggers.md` keeps its original "coming" text, because with this gating artifact triggers genuinely are not available to that audience.

One consequence worth stating plainly, since it now stays open: the 10+ generated API-reference pages for these same symbols remain `+flyte +union`, so the OSS variant still ships the artifacts API and CLI reference with no prose explaining it. Tracked in **DOC-1443**, not this PR.

## Notes on the individual fixes

**3 — do not copy the SDK's own docstring here.** It passes `accelerator="A100:8"`, which is **not a parameter of `hf_model`**. The real one is `Resources(gpu=…)`, verified against the released signature (`cpu, memory, gpu, disk, shm`). That upstream docstring is its own bug.

**4 — the box was right about the outcome and wrong about the cause.** Unwrapping and the `ProducedArtifact` declaration happen *regardless* of the flag (`test_declaring_is_unconditional`); `produces_artifacts` gates backend extraction. Reworded so a reader reasoning from it reasons correctly, since the original would mislead anyone debugging.

**6 — had to be variant-guarded, not rewritten.** `triggers.md` is `+flyte +union` and the artifacts pages are `-flyte +union`. A flat rewrite would assert a Union-only capability to OSS readers *and* create a `+flyte` page linking to a `-flyte` page, which fails `check-links`.

**7 — no CI check could have caught this.** A *missing* link is not a *broken* link, so everything stayed green while the section was unreachable from its own directory page. The card sits inside the union block deliberately: the unguarded "Basics" grid renders in the flyte variant too, and a card there pointing at a `-flyte` page is the variant-safety bug this review was watching for.

## Verification

- `check_internal_links.py` — **`flyte: all links OK` / `union: all links OK`**, run against the branch with the submodule materialized. This is what makes fix 6 safe.
- Corrected snippets **executed** against flyte `2.6.1`, not just read: `flyte.app.AppEnvironment(...)` constructs, `import flyte.prefetch` resolves, `Resources(gpu=…)` accepts.
- Added prose passes the de-slop gate: zero em-dashes, zero tell-vocabulary.
- Submodule pointers untouched.

## Still outstanding (not addressed here)

The non-blocking list from my review stands, and a few are factual rather than cosmetic — the model card is only attached *if the repo has one*, the UI tab is `Artifact card` (sentence case) and there is also an `Overview` tab, trigger rows *expand* rather than navigate, and the "CLI accepts the same options as the Python API" claim is not quite right (the CLI splits resources into individual flags; there is no `gpu=` Python kwarg). Say the word and I will push those too.

---
— docsy · automated docs agent · [DOC-1207](https://linear.app/unionai/issue/DOC-1207/docs-coverage-artifacts) · [run report](https://github.com/unionai/docsy/blob/main/runs/2026-08-18-docsy-artifacts-pr-1454.md)

ppiegaze:
## Finding 5 done — prefetch pages de-duplicated and cross-linked (`809ef12`)

Per Peeter: keep `artifacts/prefetch-models.md`, strip what the general page already covers, link both ways.

### What was duplicated, and what survived

Three of the artifacts page's four sections duplicated `apps/serve-and-deploy-apps/prefetching-models.md` — the basic example, sharding, and CLI usage. All how-to content, all covered in more depth on the general page. Removed.

**What did not duplicate is the artifact semantics, and the general page states none of it:** that a prefetch registers a model artifact at all, that the version is the Hugging Face commit (so re-prefetching an unchanged model republishes the same version instead of duplicating it), what the searchable metadata carries, that the README becomes the model card, and how to retrieve the result by name or by `hf://` source reference.

The general page's only gestures at this are a bullet saying models are "versioned and stored in Flyte's object store," a "Custom artifact name" section, and a best-practice tip about naming. So it is a real gap there, and the right subject for a page in the artifacts section.

Page is **52 lines, down from 71**. For calibration, `lineage.md` is 43 and `artifacts-in-apps.md` is 42, so it is the shortest in the section but not an outlier.

### The backlink had to be variant-guarded

The general page is `+flyte +union`, the artifacts pages are `-flyte +union`. I added an unguarded link and ran the checker to confirm rather than assuming:

```
flyte: 1 error(s)
  prefetching-models.md: page exists but not in this variant:
  ../../artifacts/prefetch-models
```

So it is wrapped in `{{< variant union >}}`, same pattern as the `triggers.md` fix. Artifacts → general is fine unguarded.

### Two CLI flags on the general page do not exist

Worth flagging separately, because it is the page the artifacts page now routes readers to for CLI usage, and it was the less accurate of the two:

| Documented | Actual |
|---|---|
| `--accelerator` | **`--gpu`** |
| `--ephemeral-storage` | **`--disk`** |

Neither appears anywhere in `flyte prefetch hf-model`'s option list. Its full set is `--architecture --artifact-name --cpu --disk --force --format --gpu --hf-token-key --mem --modality --model-type --raw-data-path --shard-config --shm --short-description --task --wait`.

The page's "Complete CLI usage" block was also missing two real flags (`--shm`, `--raw-data-path`), so that claim was false. Both corrected and added.

### One instance could not be fixed here

The same bad flag reaches `api-reference/flyte-cli.md` from the SDK's own help text (`src/flyte/cli/_prefetch.py:210` advertises `--accelerator`, which that file never defines). That page is **generated**, so the only fix is upstream: **[flyteorg/flyte-sdk#1451](https://github.com/flyteorg/flyte-sdk/pull/1451)**, which also corrects the `hf_model()` docstring that caused the original broken example in this PR.

That makes three instances of one drift: the Python docstring, the CLI help text, and this page. All three now fixed or filed.

**14/14 checks green.** `check_internal_links.py` passes on both variants.

---
— docsy · automated docs agent · [DOC-1207](https://linear.app/unionai/issue/DOC-1207/docs-coverage-artifacts) · [run report](https://github.com/unionai/docsy/blob/main/runs/2026-08-18-docsy-artifacts-pr-1454.md)

ppiegaze:
## Non-blocking items pushed (`bb36547`) — review findings now fully disposed of

Each checked against source rather than reasoned about. **14/14 green.**

| Item | Was | Now | Evidence |
|---|---|---|---|
| Lineage node behavior | "Each node links to its own page" | app/run nodes navigate; **trigger rows expand in place** | `clientsv2/.../lineage/graph.ts` — apps and runs carry `href`, trigger rows are `expandable: true` |
| UI tab label | `Artifact Card` | **`Artifact card`** | `cloud@origin/main` `ArtifactDetails/Main.tsx` |
| Filler sentence | "This gives you a birds-eye view…" | deleted | misspelled (`bird's-eye`), and the next two sentences already say it concretely |
| Casing | `commit id` | **`commit ID`** | corpus convention (`caching.md`, `links.md`, `rerun-runs.md`, `triggers.md`) |
| Apps card icon | `rocket` | **`window`** | how Apps is iconed on `user-guide/_index.md` and `tasks/_index.md`; `rocket` is the corpus's advanced-project icon |
| App-published artifacts | claim flagged unverified | **kept, caveat added** | see below |

### Two judgment calls worth stating

**I kept the "apps can publish artifacts" claim instead of dropping it.** My review flagged it as unverifiable. Checking further, the SDK's own examples document `Artifact.create(...)` as callable **"from anywhere (no task)"** and **"from any Python process"** (`examples/artifacts/artifact_example.py`, `examples/triggers/artifact_trigger_external.py`), and an app endpoint is such a process. So the capability is supported and dropping it would have removed something true and useful.

What *is* verifiable and does matter is the consequence: outside a task there is no producing run to stamp on the version, so provenance needs `external_ref`. That caveat is now on the page. The finding was right that something was missing; it was the caveat, not the claim.

**I deliberately did not add the `Overview` tab** to the tab sentence, though it exists. That sentence lists the tabs which *break the same relationships out as lists*, and Overview is not one of those. Adding it would have made the sentence less accurate, not more. Full tab set for the record: Overview, Versions, Lineage, Artifact card, Triggers, Apps.

### Where the review now stands

All seven blocking findings and every non-blocking item are disposed of. Two things remain open and are **not** in this PR by design:

- **Examples are inline rather than pulled from `unionai-examples`.** The house pattern is `{{< code file=… fragment=… >}}` against tested examples, used by 78 of 164 user-guide pages. Three of this PR's snippets shipped broken, which is exactly what that convention prevents. Worth a follow-up; `flyte-sdk/examples/artifacts/artifact_example.py` and `examples/triggers/artifact_trigger.py` are ready-made sources.
- **`flyte.TriggeredArtifact` and `flyte.TriggerTime` are in no linkmap**, so they will not autolink. Not a build break, but it suggests they are missing from the generated API reference. If they are public, that is a `--refresh-api` regen, not a docs edit.

---
— docsy · automated docs agent · [DOC-1207](https://linear.app/unionai/issue/DOC-1207/docs-coverage-artifacts) · [run report](https://github.com/unionai/docsy/blob/main/runs/2026-08-18-docsy-artifacts-pr-1454.md)

ppiegaze:
## Independent re-review, and a blocking bug it caught (`97d6f9e`)

Peeter asked for an independent re-review. It ran in a clean context with a neutral brief: it was **not** told what the earlier passes found or fixed, so it re-derived everything from the current head against shipped source. It found something three prior rounds missed, **in text I wrote**, not the contributor's.

### The bug

`artifacts/prefetch-models.md` told readers to find a prefetched model with:

```bash
flyte get artifact --source-external-ref hf://HuggingFaceTB/SmolLM2-135M-Instruct
```

**A prefetch never sets `external_ref`, so that command matches nothing.** It exits cleanly with an empty list, which reads as "no such model" rather than "wrong filter" — which is exactly why it survived review three times. A wrong *value* is not an error.

Verified independently before acting on it:

* `external_ref` is set in **one place** in the whole SDK: `Artifact.create()` (`remote/_artifact.py:287`). It appears **zero times** anywhere in `flyte/prefetch/`.
* Prefetch registers via the task-output path instead: `_hf_model.py:885` is `env.task(produces_artifacts=True)` returning `artifacts.new(...)`, and `ProducedArtifact` has no `external_ref` field.
* Backend-side the two are mutually exclusive **by construction**: `ArtifactSource` is a `oneof`, so a task-produced artifact carries `task_action` and can never also carry `external_ref`.
* What a prefetch actually records is `attrs = {"source_repo", "source_commit"}` (`_hf_model.py:598`), bare repo id, no scheme.

### The fix, verified as hard as the finding

```bash
flyte get artifact --attr source_repo=HuggingFaceTB/SmolLM2-135M-Instruct
```

`--attr` maps to `Filter(VALUE_IN, field=user_metadata.<key>)` (`remote/_artifact.py:407-418`), which the server supports for arbitrary `user_metadata` keys. Checked before swapping it in, rather than trading one plausible-looking command for another.

Also dropped the lineage cross-link in that sentence: for a prefetch the graph's source node is the **prefetch run**, not an external source. The "External source / Imported" node renders only for the `externalRef` case, which a prefetch never takes.

**The tell I missed:** the PR contradicted itself. `lineage.md` says a prefetch "records the source **repo and commit**", and the backlink I added to the general page promises retrieval "by **source repo**". Only this one spot said external-ref, and it happened to be the spot with a runnable command in it.

### Two one-line corrections in the same commit

* **`lineage.md`** — the graph is not uniformly version-scoped. Apps are queried by `(name, version)`, **triggers by name alone** (`ArtifactLineageTab.tsx:46-51`; `graph.ts:150` says so outright).
* **`task-outputs.md`** — `create_model_metadata` does **not** set the `Metadata.kind` field. It leaves `kind` as `None` and stamps `attrs["flyte.io/kind"] = "model"` (`_metadata.py:68-86`). Confirmed at runtime: `m.kind` is `None`. Reworded so a reader who inspects the field is not surprised.

### What the independent pass confirmed

Worth recording, since a review that only lists defects gives a skewed picture. It executed the trigger and apps snippets against **flyte 2.6.2** (both construct); confirmed the corrected `produces_artifacts` WARNING box **including its mechanism**; confirmed the multi-output example is genuinely top-level per `convert.py:415-417`; confirmed all three trigger "Rules" map one-to-one onto `_trigger.py:830-844`; confirmed `model_path: str | RunOutput | ArtifactValue` from the **installed** vLLM/SGLang plugin packages rather than from generated docs; confirmed both CLI sections match `--help` flag for flag; confirmed the six console tabs and the expandable-trigger behavior at `cloud@origin/main`; and re-ran `check_internal_links.py` on a worktree of the head — `flyte: all links OK / union: all links OK`.

It also noted my brief was one release stale (it grounded against **2.6.2**, not the 2.6.1 I cited).

**Its recommendation was merge-after-one-fix.** That fix is in. **14/14 green.**

### Still open, unchanged

The variant split (this guide `-flyte +union` while the artifacts API reference is `+flyte +union`) is a settled call tracked in **DOC-1443** / **DOC-1445**. `flyte.TriggeredArtifact` and `flyte.TriggerTime` remain absent from every linkmap, so they will not autolink; they are module-level singleton *instances*, which is likely why the generator skips them, and that is a regen question rather than a docs edit.

---
— docsy · automated docs agent · [DOC-1207](https://linear.app/unionai/issue/DOC-1207/docs-coverage-artifacts) · [run report](https://github.com/unionai/docsy/blob/main/runs/2026-08-18-docsy-artifacts-pr-1454.md)

PR #1463, opened 2026-08-19
docsy(v2): document Ray autoscaler_options (AutoscalerOptionsConfig)

ppiegaze:
## The gap

`content/integrations/ray/_index.md` documented `enable_autoscaling`, the on/off switch, and nothing that configures it. No `autoscaler_options` row, no mention of `AutoscalerOptionsConfig` anywhere on the page. So the page taught the feature while omitting every knob.

The generated reference already carried the class (`content/api-reference/integrations/ray/autoscaleroptionsconfig.md`, added by regen #1442); only the narrative lagged. Worth noting the generated parameter table has **empty descriptions**, because the generator does not parse the docstring's field list, so the narrative table is not redundant with it.

## Why this was missed

#1442 spun [DOC-1442](https://linear.app/unionai/issue/DOC-1442), which covered only `flyte create config --devbox` (shipped as #1453). This delta was never dispositioned, and #1442 was then omitted from [#1459](https://github.com/unionai/unionai-docs/pull/1459)'s scope without explanation. An independent review of #1459 caught it by enumerating the full `docsy:api-surface-delta` label set: **#1262, #1283 (v1), #1373, #1424, #1442, #1457**, five regens rather than the four #1459 claimed.

## Verified against the released tag

New in **v2.6.1**: `autoscaler_options` has **0** occurrences in `v2.6.0:plugins/ray/src/flyteplugins/ray/task.py` and **4** in `v2.6.1`. `AutoscalerOptionsConfig` is exported from `flyteplugins.ray.__init__` and is in `__all__`, so the example's import line works as written.

Three behaviors are documented because none is inferable from the signature:

| Behavior | Source |
|---|---|
| **`autoscaler_options` does not enable autoscaling.** The Go plugin sets `EnableInTreeAutoscaling` and `AutoscalerOptions` as independent fields on the RayCluster spec, and KubeRay creates the sidecar only when in-tree autoscaling is on, so options alone do nothing | `flyteorg/flyte:flyteplugins/go/tasks/plugins/k8s/ray/ray.go:267-268` |
| **Unset fields keep KubeRay defaults rather than being zeroed** — `idle_timeout_seconds` applies only when `> 0`, `upscaling_mode` only when not `UNSPECIFIED` | `ray.go:208-236` `buildAutoscalerOptions` |
| **`resources` takes tuples** for request/limit pairs | `task.py:36-59` docstring, `get_proto_resources` |

Upscaling mode is documented with the **SDK spelling** (`AutoscalerOptionsConfig.UpscalingMode.CONSERVATIVE`), not the KubeRay string `"Conservative"` that the plugin maps it to.

## Sequencing — must rebase before merge

[#1458](https://github.com/unionai/unionai-docs/pull/1458) is editing **the same section of this page** (its head-node resource block inserts immediately after the `HeadNodeConfig` table, which is exactly where this section begins). A conflict is expected and it is a trivial add-vs-add. **Merge #1458 first, then rebase this.**

## Not addressed

`buildAutoscalerOptions` swallows a `Resources` conversion error (`if res, err := ...; err == nil`), so an invalid `resources` silently produces no resources rather than an error. Left undocumented pending a view on whether that is intended behavior or a bug worth an eng finding.

## Checks

`make dist` green both variants, `make check-links` green, `make check-icon-names` green, markdownlint 0 issues.

---
— docsy · automated docs agent · [DOC-1450](https://linear.app/unionai/issue/DOC-1450)

### DIFF
--- a/content/integrations/ray/_index.md
+++ b/content/integrations/ray/_index.md
@@ -76,6 +76,7 @@ ray_env = flyte.TaskEnvironment(
 | `worker_node_config` | `List[WorkerNodeConfig]` | **Required.** List of worker group configurations |
 | `head_node_config` | `HeadNodeConfig` | Head node configuration (optional) |
 | `enable_autoscaling` | `bool` | Enable Ray autoscaler (default: `False`) |
+| `autoscaler_options` | `AutoscalerOptionsConfig` | Tune the autoscaler sidecar. Has no effect unless `enable_autoscaling` is `True` |
 | `runtime_env` | `dict` | Ray runtime environment (pip packages, env vars, etc.) |
 | `address` | `str` | Connect to an existing Ray cluster instead of provisioning one |
 | `shutdown_after_job_finishes` | `bool` | Shut down the cluster after the job completes (default: `False`) |
@@ -117,6 +118,39 @@ Under-provisioning the head node is a common cause of a cluster that never becom
 ready: if the dashboard cannot start, `ray start --head` fails and KubeRay recycles
 the head pod in a loop. A head pod at 1 CPU and 1000Mi has been observed failing
 this way.
+
+### `AutoscalerOptionsConfig` parameters
+
+Setting `enable_autoscaling=True` runs the Ray autoscaler with KubeRay's defaults. Pass `autoscaler_options` to tune it:
+
+```python
+from flyteplugins.ray import AutoscalerOptionsConfig, RayJobConfig, WorkerNodeConfig
+
+ray_config = RayJobConfig(
+    worker_node_config=[
+        WorkerNodeConfig(group_name="ray-group", replicas=1, min_replicas=1, max_replicas=5)
+    ],
+    enable_autoscaling=True,
+    autoscaler_options=AutoscalerOptionsConfig(
+        upscaling_mode=AutoscalerOptionsConfig.UpscalingMode.CONSERVATIVE,
+        idle_timeout_seconds=120,
+        resources=flyte.Resources(cpu=("500m", "1"), memory=("512Mi", "1Gi")),
+    ),
+)
+```
+
+| Parameter | Type | Description |
+|-----------|------|-------------|
+| `upscaling_mode` | `AutoscalerOptionsConfig.UpscalingMode` | How aggressively to add nodes. One of `DEFAULT`, `AGGRESSIVE`, `CONSERVATIVE`, `UNSPECIFIED` |
+| `idle_timeout_seconds` | `int` | Seconds a node may sit idle before the autoscaler removes it |
+| `image` | `str` | Container image for the autoscaler sidecar |
+| `env` | `Dict[str, str]` | Environment variables for the autoscaler container |
+| `resources` | `Resources` | Requests and limits for the autoscaler sidecar |
+
+Every field is optional, and any field you leave unset keeps the KubeRay default rather than being zeroed. `resources` accepts tuples to set a request and a limit together, as in `flyte.Resources(cpu=("500m", "1"))`.
+
+> [!NOTE] The options do not switch autoscaling on
+> `autoscaler_options` only configures the autoscaler sidecar, and the sidecar is created only when `enable_autoscaling` is `True`. Passing options on their own changes nothing.
 
 ### Connecting to an existing cluster
 
--- a/content/integrations/ray/_index.md
+++ b/content/integrations/ray/_index.md
@@ -141,13 +141,13 @@ ray_config = RayJobConfig(
 
 | Parameter | Type | Description |
 |-----------|------|-------------|
-| `upscaling_mode` | `AutoscalerOptionsConfig.UpscalingMode` | How aggressively to add nodes. One of `DEFAULT`, `AGGRESSIVE`, `CONSERVATIVE`, `UNSPECIFIED` |
+| `upscaling_mode` | `AutoscalerOptionsConfig.UpscalingMode` | Rate limiting on adding nodes. `CONSERVATIVE` holds the number of pending worker pods to at most the current cluster size. `DEFAULT` and `AGGRESSIVE` are the same setting: no rate limit. Leaving the field unset, or passing `UNSPECIFIED`, also means no rate limit |
 | `idle_timeout_seconds` | `int` | Seconds a node may sit idle before the autoscaler removes it |
 | `image` | `str` | Container image for the autoscaler sidecar |
 | `env` | `Dict[str, str]` | Environment variables for the autoscaler container |
 | `resources` | `Resources` | Requests and limits for the autoscaler sidecar |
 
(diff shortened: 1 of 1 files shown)


content/integrations/ray/_index.md
```diff
@@ -94,6 +95,39 @@ ray_env = flyte.TaskEnvironment(
...
+
+| Parameter | Type | Description |
+|-----------|------|-------------|
+| `upscaling_mode` | `AutoscalerOptionsConfig.UpscalingMode` | How aggressively to add nodes. One of `DEFAULT`, `AGGRESSIVE`, `CONSERVATIVE`, `UNSPECIFIED` |
+| `idle_timeout_seconds` | `int` | Seconds a node may sit idle before the autoscaler removes it |
+| `image` | `str` | Container image for the autoscaler sidecar |
+| `env` | `Dict[str, str]` | Environment variables for the autoscaler container |
+| `resources` | `Resources` | Requests and limits for the autoscaler sidecar |
+
+Every field is optional, and any field you leave unset keeps the KubeRay default rather than being zeroed. `resources` accepts tuples to set a request and a limit together, as in `flyte.Resources(cpu=("500m", "1"))`.
```
ppiegaze: **blocking.** False for `resources`. Passing `autoscaler_options` at all sets `AutoscalerOptions.Resources` to a non-nil empty value, and KubeRay replaces its 500m CPU / 512Mi request and limit defaults wholesale rather than merging them. So omitting `resources`, or setting only a request, leaves the sidecar without them. (`pod.go:730`)

```suggestion
Every field is optional. Leaving `upscaling_mode`, `idle_timeout_seconds`, `image`, or `env` unset keeps the KubeRay default. `resources` is the exception: whenever you pass `autoscaler_options`, whatever you give for `resources` replaces the sidecar's default 500m CPU and 512Mi memory requests and limits outright. Omit it and the sidecar runs with no requests or limits at all; set only a request and it also loses the default limits. Set `resources` explicitly, on both sides. It accepts tuples to set a request and a limit together, as in `flyte.Resources(cpu=("500m", "1"))`.
```

ppiegaze (commented):
Reviewed at `48239d69` against flyte-sdk `v2.6.3`, `flyteorg/flyte@main` and kuberay `v1.6.2`. Types, import path and the "options do not switch autoscaling on" callout all check out; 4 changes needed, 2 blocking.

| # | Finding | Where | Blocking |
|---|---|---|---|
| F1 | "unset keeps the KubeRay default" is false for `resources`: the sidecar loses its 500m/512Mi requests and limits | L126 | yes |
| F2 | `DEFAULT` and `AGGRESSIVE` are the same mode; `UNSPECIFIED` is a sentinel, not a choice | L120 | yes |
| F3 | `idle_timeout_seconds` default of 60 unstated; an explicit `0` is dropped | L121 | no |
| F4 | Snippet declares imports but uses `flyte.Resources` without importing `flyte` | L103 | no |
| F6 | Publish gate: backend rollout not confirmed, only merged into `cloud@main`'s `flyte2` pointer | n/a | escalated |

---
— docsy · automated docs agent · [DOC-1450](https://linear.app/unionai/issue/DOC-1450)

content/integrations/ray/_index.md
```diff
@@ -94,6 +95,39 @@ ray_env = flyte.TaskEnvironment(
...
+        upscaling_mode=AutoscalerOptionsConfig.UpscalingMode.CONSERVATIVE,
+        idle_timeout_seconds=120,
+        resources=flyte.Resources(cpu=("500m", "1"), memory=("512Mi", "1Gi")),
+    ),
+)
+```
+
+| Parameter | Type | Description |
+|-----------|------|-------------|
+| `upscaling_mode` | `AutoscalerOptionsConfig.UpscalingMode` | How aggressively to add nodes. One of `DEFAULT`, `AGGRESSIVE`, `CONSERVATIVE`, `UNSPECIFIED` |
```
ppiegaze: **blocking.** These are not four graded settings. Ray maps `DEFAULT` and `AGGRESSIVE` to the same `upscaling_speed` of 1000 and marks the second branch redundant in its own source. `UNSPECIFIED` is a sentinel the plugin drops before the CR, so it is not a mode a reader can pick. Only `CONSERVATIVE` behaves differently. (`autoscaling_config.py:110`)

```suggestion
| `upscaling_mode` | `AutoscalerOptionsConfig.UpscalingMode` | Rate limiting on adding nodes. `CONSERVATIVE` holds the number of pending worker pods to at most the current cluster size. `DEFAULT` and `AGGRESSIVE` are the same setting: no rate limit. Leaving the field unset, or passing `UNSPECIFIED`, also means no rate limit |
```

ppiegaze:
**nit** (non-blocking), 2 items.

**L121** `idle_timeout_seconds` has a KubeRay default of 60 seconds that the row does not state, and an explicit `0` is dropped by the plugin's `> 0` guard rather than disabling idle removal. Suggested cell text: `Seconds a node may sit idle before the autoscaler removes it (default: 60). An explicit 0 is dropped and the default applies`.

**L103** The block opens with an import line, which advertises a self-contained snippet, but `flyte.Resources` at L113 is never imported. Add `import flyte` above L103, matching the snippet at L150.

content/integrations/ray/_index.md
ppiegaze: Applied in `cc067b1c`. Re-verified before applying: `ToK8sResourceRequirements(nil)` returns a non-nil empty struct with a nil error (`utils.go:53-55`), so the `err == nil` guard cannot fail and KubeRay takes the replace branch at `pod.go:730-731`.

content/integrations/ray/_index.md
ppiegaze: Applied in `cc067b1c`. Confirmed at `autoscaling_config.py:107-116`: `CONSERVATIVE` maps to `upscaling_speed` 1, the other three to 1000, and Ray calls the `AGGRESSIVE` branch redundant in its own comment.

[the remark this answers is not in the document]
ppiegaze: Covered by the rewritten row in `cc067b1c`: `UNSPECIFIED` is now described as equivalent to leaving the field unset, not as a selectable mode.

content/integrations/ray/_index.md
```diff
@@ -94,6 +95,39 @@ ray_env = flyte.TaskEnvironment(
...
 | `pod_template` | `PodTemplate` | Full pod template (mutually exclusive with `requests`/`limits`) |
 
+### `AutoscalerOptionsConfig` parameters
+
+Setting `enable_autoscaling=True` runs the Ray autoscaler with KubeRay's defaults. Pass `autoscaler_options` to tune it:
+
+```python
+from flyteplugins.ray import AutoscalerOptionsConfig, RayJobConfig, WorkerNodeConfig
+
+ray_config = RayJobConfig(
```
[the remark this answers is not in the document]
ppiegaze: Applied in `81375824`. Also states the `idle_timeout_seconds` default of 60 (`raycluster_types.go:226-228`, kuberay v1.6.2) and that an explicit `0` is dropped by the `> 0` guard at `ray.go:140` rather than disabling idle removal.

PR #1481, opened 2026-08-20
docsy(v2): drive the v1 API-docs regen from the default branch

ppiegaze:
[DOC-1293](https://linear.app/unionai/issue/DOC-1293), open since 2026-07-29 as *"a follow-up, not wired yet."*

## The constraint

GitHub fires `schedule` and `repository_dispatch` **only from the default branch**. `regen-api-docs.yml` exists on `v1` and works — but its copy there can only declare `workflow_dispatch`, because a `schedule:` in a v1-branch file would never run.

So v1 never auto-regenerated. Its API reference sat at flytekit **1.16.26** while **1.16.28** was released, and moved only when a human remembered to dispatch it by hand — which, verified before this PR, nobody ever had.

## Deliberately thin: dispatch, don't reimplement

The driver does **not** duplicate the regen. It dispatches the existing v1 workflow with `ref: v1`, so v1 keeps **one** implementation of its own regen — drift gate, version fold, PR body, concurrency group.

Duplicating that logic here would give the two lines two regens to keep in sync, and the v1 copy is the one nobody looks at. That is precisely the failure mode `LESSONS.md` §48 is about, and this repo has already paid for it three times today (the inert `push:` filter on v1's icon check, the `versions.toml` promote, the missing auto-regen itself).

## Why the App token

**A workflow dispatched with `GITHUB_TOKEN` does not create a run** — GitHub suppresses event cascades from that token to prevent loops. The `unionai-docsy-bot` App is already configured here (it authors the regen PRs, e.g. #1477), so the driver mints an App token exactly as the regen does.

## Both failure paths are loud

| Failure | Behaviour |
|---|---|
| App secrets missing | Job **fails** with a specific error, rather than dispatching into silence |
| Dispatch call fails | Exits non-zero |

A driver that quietly does nothing is **worse than no driver**: v1 would be exactly as stale while looking handled. That is the same shape as the swallowed `eval` fixed in #1478/#1479 earlier today, and it seemed worth not rebuilding it here.

## Safe to run daily

Everything downstream of the v1 regen's `drift` gate is conditioned on it (`Regenerate`, `Fold`, `Classify`, `Build PR body`, `Create draft PR` — all `if: steps.drift.outputs.drift == 'true'`). A run with no PyPI drift regenerates nothing and opens no PR; the cost of a quiet day is a few CI minutes.

Scheduled **13:30 UTC**, 30 minutes after main's own regen at 13:00, so the two lines do not contend for runners or land two regen PRs in the same minute.

## ⚠️ Deployment prerequisite

**Dispatching a workflow requires `Actions: write` on the App installation.** The App is documented in `regen-api-docs.yml`'s header as installed with *Contents: RW + Pull requests: RW*, so that scope likely needs adding.

If it is missing, the first scheduled run **fails visibly** — which is the intended behaviour, but worth granting up front. Recommend running the driver once via `workflow_dispatch` after merge to confirm, rather than waiting for 13:30 UTC to find out.

## Not done here

A flytekit-release → `repository_dispatch` signal, the analogue of flyte-sdk's for v2. That needs a PR against the flytekit repo to emit the dispatch, which is outside this repo's control. The daily poll covers the same ground with at most ~24h latency — what main's own schedule already calls a *"daily backstop poll."* Worth adding later only if that latency ever matters.

---
— docsy · automated docs agent · [DOC-1293](https://linear.app/unionai/issue/DOC-1293)

### DIFF
--- a/.github/workflows/regen-api-docs-v1-driver.yml
+++ b/.github/workflows/regen-api-docs-v1-driver.yml
@@ -1,0 +1,91 @@
+# Drive the v1 API-docs regen from the DEFAULT branch (DOC-1293).
+#
+# THE CONSTRAINT: GitHub fires `schedule` and `repository_dispatch` only from the
+# repo's default branch. `regen-api-docs.yml` exists on `v1` and works, but its
+# copy there can only declare `workflow_dispatch` — a `schedule:` on a v1-branch
+# file would never run. So v1 never auto-regenerated: its API reference sat at
+# flytekit 1.16.26 while 1.16.28 was released, and moved only when a human
+# remembered to dispatch it by hand.
+#
+# THE FIX, deliberately thin: this driver does NOT reimplement the regen. It
+# dispatches the existing v1 workflow with `ref: v1`. v1 keeps ONE implementation
+# of its own regen — its drift gate, its fold, its PR body, its concurrency group.
+# Duplicating that logic here would give the two lines two regens to keep in sync,
+# and the v1 copy is the one nobody looks at (LESSONS.md §48).
+#
+# WHY THE APP TOKEN: a workflow dispatched using GITHUB_TOKEN does not create a
+# new workflow run — GitHub suppresses event cascades from that token to prevent
+# loops. The docsy-bot App is already configured here (it authors the regen PRs),
+# so the driver mints an App token exactly as the regen does. Without the App
+# secrets this job fails loudly rather than dispatching into silence, because a
+# driver that quietly does nothing is worse than no driver: it looks handled.
+#
+# SAFE TO RUN DAILY: everything downstream of the v1 regen's `drift` gate is
+# conditioned on it, so a run with no PyPI drift regenerates nothing and opens no
+# PR. The cost of a quiet day is a few CI minutes.
+#
+# NOT DONE HERE: a flytekit-release -> repository_dispatch signal, the analogue of
+# flyte-sdk's for v2. That needs a PR against the flytekit repo to emit the
+# dispatch, which is out of this repo's control. The daily poll covers the same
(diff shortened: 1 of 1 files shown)


ppiegaze:
## Confirmed: the App cannot dispatch yet — `actions` is absent

Queried the App's declared permissions directly (`GET /apps/unionai-docsy-bot`, public metadata):

```json
{
  "slug": "unionai-docsy-bot",
  "permissions": {
    "contents": "write",
    "issues": "write",
    "metadata": "read",
    "pull_requests": "write"
  }
}
```

**No `actions` key.** `gh workflow run` requires **`actions: write`**, so as things stand this driver's dispatch step will fail on its first run. Loudly and with a clear message, which is the intended design — but it will not work until the scope is granted.

This is the App's *declared* permission set, which is an upper bound: an installation can never hold more than the App declares, so no installation-level check can change the answer.

## To unblock, two steps

1. **App settings → Permissions & events → Repository permissions → Actions: Read and write.**
   `https://github.com/organizations/unionai/settings/apps/unionai-docsy-bot/permissions`
2. **Approve the permission change on the installation.** GitHub does not apply a widened scope silently — it notifies org owners and the installation must accept it. Until accepted, the token still carries the old set.

Then run this workflow once via `workflow_dispatch` to confirm end to end, rather than waiting for 13:30 UTC.

## Worth noting while you are in there

The App also holds **`issues: write`**, which `regen-api-docs.yml`'s header does not mention (it documents *Contents: RW + Pull requests: RW*). Nothing docsy runs appears to need it. Not urgent, but if it is unused it is a scope worth dropping while you are editing the permission set — and the header is worth correcting either way, since it is now wrong in both directions.

---
— docsy · automated docs agent · [DOC-1293](https://linear.app/unionai/issue/DOC-1293)

PR #1505, opened 2026-08-25
Update command to include TUI option for local run

LeonKolyang:
The example to launch the tui doesn't actually launch the tui, because it's missing `--tui`. Adding it here.


cosmicBboy (approved):

cosmicBboy:
need to fix DCO @LeonKolyang

ppiegaze (approved):

